PR #3, opened 2018-11-17
Add "Abandoned Isles" map by gimp

mrowlbio:
I've changed the initial stuff for this map to a stone sword a stone pick and three torches.

Here are some screenshots of it:

EDIT: Changed the screenshots to default textures


09_abandoned_isles.conf
```diff
@@ -0,0 +1,6 @@
+r = 115
+name = abandoned_isles
+h = 140
+rotation = z
+author = gimp
+initial_stuff = default:sword_stone,default:pick_stone
```
ClobberXD: Missing conf keys:
```
team.1 = red
team.1.color = red
team.1.pos = <pos of flag 1>
team.2 = blue
team.2.color = blue
team.2.pos = <pos of flag 2>
```

This issue is most likely because an old version of `ctf_map` has been used, as the missing conf values issue was fixed more than 2 months ago, in https://github.com/MT-CTF/capturetheflag/commit/6375bb2cf0cf3da23c692bdd9dca5ce201fdf9c9

ClobberXD: I'll see if I can manually add it to the conf...

mrowlbio: Ok, I will try to get it working. Last time I tried to add flag positions it put the spawn in the wrong position everytime.

MinetestSam:
You can manually add the coords of the flags position to the config file

09_abandoned_isles.conf
ClobberXD: That's because the coords are relative to the center of the schem. This is because it's a schem, and the flag positions are bound to change depending on its placement.

MinetestSam: You can manually add the coords of the flags position to the config file

MinetestSam:
The schem provided does cut off. Barriers are missing 
screenshot_20181117_214104
Also if you don't want the sand to fall off then just go to the properties of sand and turn off falling_node, i.e, falling_node = 0, then you wouldn't need to place cobble below the sand cube

mrowlbio:
Ive been adding some more islands and made the cubes more natural looking with the power of tnt, and exported it properly

09_abandoned_isles.conf
mrowlbio: Done, I tested it on LAN and it worked

afshaan4:
Gimp and I are going to applying the feedback about making the cubes more natural that was given on the discord, so don't merge yet

mrowlbio:
I finally finished it. I have made the big islands pointy as requested and have added a lava floor also I have added tiny indestructible islands to stop the water from reaching till the lava and have also made critical areas indestructible as ANAND suggested.

mrowlbio:
ANAND said some more land would be nice, and I have obliged. Demo video is on #maps on discord.

09_abandoned_isles.conf
```diff
@@ -0,0 +1,12 @@
...
+team.1 = red
+team.1.pos = (-80,-1,91)
+team.2.pos = (85,-1,-101)
+team.2.color = blue
+author = gimp and caoutiouspotato
+rotation = x
+h = 140
+team.2 = blue
+name = Abandoned Isles
+initial_stuff = default_sword:steel,shooter:pistol,default:sandstone 50
```
Thomas--S: Are you sure this is correct?

```suggestion
initial_stuff = default:sword_steel,shooter:pistol,default:sandstone 50
```

mrowlbio: Whoops, fixed it.

BrianGaucher:
I decided to review the map. Remember, this is only my opinion, and my review will be subjective. My opinion is not final either.

The map is very wide, but since it is quite lightly populated, it works fine. So good size.

*Water cups*: The falling water is nice, having "cups" to catch the water is cool, but it's still possible to overflow it by swimming down the water then blocking up the hole, overflowing the water, which can form obsidian. Players could harvest it to fortify their base, or build to the other team. I guess this could be a feature as well.
Some chest will spawn on the water "cups".
overflow

*Indestructible nodes*: On some islands not a single node can be destroyed, except for the trees. IMO, only the bottom layer should be indestructible. Some islands however are fully destructible. One island has the top (dry grass) and bottom layer(stone) indestructible, but the middle layer (dirt) is destructible. A few islands are mostly indestructible, except for the ores. 
The sandstone bricks around the flag are indestructible, with some steel doors, the only way in for the other team is through the roof.

*Ores*: One team has 9 mese, the other team has 8. 
Good amount of iron ore. Consider adding a furnace to the villages. Since steel tools can spawn in chest. Consider adding some diamond ore.

*Snow and ice*: They seem out of place, could be replaced with simple island to improve consistency. I'd even say that green grass seems out of place.
A few jumps from the pre-built paths (especially up to the sphere dirt planet) are 2 long, these could be hard on mobile (obviously they can build a bridge). 

*No apples*: Hiding a few apples in the trees could add a little bit of ressource gathering. If you attach 1 apple to each tree, that's 19 apples per team (+chest).
I also noticed that one of the main bases has a large cave on the back of it. Consider giving one to both teams.

**Less subjective comments.**
The lava at the bottom is only 2 nodes deep, should be deeper.
Not a lot of material accessible to build with.

**My suggestion**: Remove ice and snow. Improve consistency for indestructible nodes. Stone, dirt, grass, sandstone is always destructible. Desert stone, and sand is always indestructible. Make only the bottom layer of islands indestructible. Add an apple or two on some trees. Make players spawn with a few nodes.

ClobberXD:
> Make only the bottom layer of islands indestructible.

That's not a good idea, as the upper layers can then be dug by players. The base will then be floating, which makes the map potentially unplayable, as players will immediately fall to their deaths when they get out of their base.

BrianGaucher:
> > Make only the bottom layer of islands indestructible.
> 
> That's not a good idea, as the upper layers can then be dug by players. The base will then be floating, which makes the map potentially unplayable, as players will immediately fall to their deaths when they get out of their base.

I meant the bottom layer of *each* island, not only the bottom ones. I completely agree that players shouldn't be able to fall to their death. Sorry for the misunderstanding, French is my maternal language.

ClobberXD:
> I meant the bottom layer of each island, not only the bottom ones

Yes, I already understood that, and that's precisely what is a bad idea. Islands generally have a shape resembling an inverted cone/pyramid. Making the bottom layer, i.e. the layer with the least area, indestructible, means that
- The effective walkable area of the island is drastically reduced
- And the base is quite high in the air, with respect to the lower surface of the island.

Making the top layer indestructible means that the island is essentially indestructible as a whole, but is also safe from the two issues stated above.

BrianGaucher:
>  Make only the bottom layer of islands indestructible.
Sorry, I miswrote my sentence. I meant: 
> Make only the bottom layer of *each* island indestructible.
I definetely don't want players falling from spawn.

BrianGaucher:
Thank you for your opinion. I think your idea is better.
However, some islands are: dirt_with_grass / dirt / stone, with some ores. The grass and stone are indestructible, but the grass and ores are destructible. Which looks bad, creates holes in the base, and is inconsistent.

afshaan4:
>Add an apple or two on some trees.

Gimp and i decided not to have apples on the map.

MinetestSam:
There is no point in having apples since they are going to be removed in the future with the addition of medkits.

BrianGaucher:
I doubt players would have much use for apples anyways, dying from lava will be very common. I agree with your decision. Remember my review doesn't completely reflect my opinion.

@MinetestSam Will apples be fully replaced by medkits, or partially replaced? I guess medkits make more sense, especially spawning in chest. But Apples allow the map creator to hide health on the map. My map  #16 is Apple enriched.

MinetestSam:
See https://github.com/MT-CTF/capturetheflag/pull/203

ClobberXD:
Tested out the map. Much better than before, but there's still scope for improvement:

- The smaller islands can be destroyed down to the last bit of stone. It's fine to make the upper layers diggable as they're trivial/minor islands, but I think at least the bottom layer can be made indestructible. Being able to completely destroy islands could potentially cause chaos with griefers online.
- The paths connecting the larger islands are breakable, and if someone does manage to destroy the path, the map becomes unplayable, as it's not an easy job to bridge two islands across a massive lava chasm without dying.
- There's very little land-mass, and the corners of the map aren't utilized well. You could make the map smaller, add more islands, or make the existing islands bigger.
- Since there is very little land-mass, there's a severe shortage of chests. I suggest adding custom chest zones covering the flag island, and the large island, on each side. See [09_swamplands.conf](/MT-CTF/maps/blob/master/09_swamplands.conf) for an example. It's currently hard to add chest-zones manually though, sorry about that.
- The number in the file names need to be changed from 09 to 11 :)

MinetestSam:
Maybe add cobble in initial items?

afshaan4:
>The paths connecting the larger islands are breakable, and if someone does manage to destroy the path, the map becomes unplayable, as it's not an easy job to bridge two islands across a massive lava chasm without dying.

I see,we'll change that

mrowlbio:
1. I want the smaller islands to be destructible because ores are in them, only key islands I want indestructible.

2. Okay, I will make it indestructible.

3. The idea is for very little island so there are more bridges and less usual run around a normal full ground thing.

4. Ok

mrowlbio:
I will make the bridges indestructible and change the name, but why do we need more chests if the players already spawn with sufficient resources? If you think it still needs more resources then I will try to add custom chest zones.

ClobberXD:
Apart from the decent starting kit, players can find good stuff in the chests like guns, grenades, etc. If there are very few chests, only a couple of players get to enjoy the loot, which would give them a significant edge over the others. As you said, the decent started kit does alleviate this to a certain extent, but it'd still be fairly unbalanced.

BrianGaucher:
Question, will chest spawn on the water-catching island?

I agree setting custom chest zones is a good idea.
If you are setting custom chest zones. I'm guessing it will be the main island. But I'd suggest adding one on those ice islands, given they aren't connected to the path, and we're partially added just to fill up the void. Giving them extra chest could give them more importance.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

mrowlbio:
I have fixed the bridge problem. I can't seem to understand how the custom chest zones work, can someone help me?

ClobberXD:
Custom chest zones are basically cuboidal regions, which can be configured by providing the _relative_ coords of 2 diagonally opposite corners.

mrowlbio:
i deleted my fork soooooo, theres a new pull request #26

PR #6, opened 2018-11-29
Add Carpathian plains map

Extex101:
screenshot_20181129_130504
screenshot_20181129_130524
Not sure how if i placed the flags right


09_carpathian_plains_map.conf
```diff
@@ -0,0 +1,11 @@
...
+author = Extex
+hint = Find the center
+rotation = x
+r = 115
+h = 140
+initial_stuff = default:pick_wood 1, shooter:crossbow 1, shooter:arrow_black 1, shooter:arrow_white 20, default:torch 20
+team.1 = black
+team.1.color = black
+team.2 = gold
+team.2.color = gold
```
Thomas--S: I suggest to use red and blue as team colors.

If I remember correctly, red and blue are hard-coded in some places in capturetheflag.

Extex101: Ok... fixed

09_carpathian_plains_map.conf
```diff
@@ -0,0 +1,11 @@
+name = Carpathian plains
+author = Extex
+hint = Find the center
+rotation = x
+r = 115
+h = 140
+initial_stuff = default:pick_wood 1, shooter:crossbow 1, shooter:arrow_black 1, shooter:arrow_white 20, default:torch 20
```
ClobberXD: Please remove the spaces after commas

09_carpathian_plains_map.conf
```diff
@@ -0,0 +1,11 @@
+name = Carpathian plains
+author = Extex
+hint = Find the center
```
ClobberXD: The hint isn't necessary and is only required when the map can't be figured out on it's own. And what does "Find the center" mean anyway? :)

09_carpathian_plains_map.conf
ClobberXD: Also, any specific reason for using black arrows?

09_carpathian_plains_map.conf
```diff
@@ -0,0 +1,11 @@
...
+rotation = x
+r = 115
+h = 140
+initial_stuff = default:pick_wood 1, shooter:crossbow 1, shooter:arrow_black 1, shooter:arrow_white 20, default:torch 20
+team.1 = red
+team.1.color = red
+team.2 = blue
+team.2.color = blue
```
ClobberXD: Missing flag pos fields `team.1.pos` and `team.2.pos`. Note that these aren't absolute position, but the flags' distance from the centre of the map, which is complicated to calculate manually. It's recommended to use ctf_map's builtin map maker for this purpose.

ClobberXD:
This seems to be the flattest map ever, flatter than even POTP which has a relatively more undulated terrain, so this map is quite unique, IMO. However, the initial items seem to be unbalanced at first glance:
- wooden pick
- crossbow
- one black arrow
- 20 regular (white) arrows
- 20 torches

Extex101:
Oh just realized a bunch of things to fix
Map contains
 - walls
 - fire 
 - straw block
 - signs
(in little shrine top left of second pic)
Should I remove the shrine or just the extra blocks? 
*Shrine made to add something fancy to the other side of the map*

09_carpathian_plains_map.conf
Extex101: Makes players want to use crossbows
Usually players just discard crossbows so (I hope) it will make them want to use the crossbow

Extex101:
Shrine containing lava (removable if not wanted)
screenshot_20181129_205556
Because the other side has the cool Carpathian weirdness it needed some coolness balance
(But if you don't want it i'd be fine to remove it)
 Also how do i chose flag positions?

Extex101:
@ClobberXD or @rubenwardy how do I set the flag locations

ClobberXD:
Can't set flag locations after creating the map. That's a serious limitation of the map maker in its current state. It'll have to be manually calculated. I could do that for you.

Extex101:
k thanks

ClobberXD:
Around 15% of the schem has been clipped off:
screenshot_20181219_142631

ClobberXD:
@Extex101 The schem is cut-off - would you be able to fix this? Are you still interested in working on this?

Extex101:
Yeah ill do it when I have a chance

ClobberXD:
We already have plain_battle which is a also very flat map (albeit with much more man-made structures). With the exception of existing maps, maps that are very similar to existing ones won't be approved. So, closing.

PR #7, opened 2018-12-26
Add Four Tunnels map

chkno:
This is my first time creating a CTF map; I'm not sure I'm doing it correctly.

I wasn't able to bring up a local CTF server to test it in CTF mode (the closest I got was "ERROR[Main]: ServerError: AsyncErr: ServerThread::run Lua: Runtime error from mod '*builtin*' in callback on_joinplayer(): .../.minetest/mods/ctf_map/give_initial_stuff.lua:32: attempt to call global 'give_initial_stuff' (a table value)").

I don't know how to add chests with additional items (this is not covered in https://github.com/MT-CTF/capturetheflag/blob/master/mods/ctf_map/README.md ), so there are no chests.

I'm happy to make any suggested edits.


09_four_tunnels.conf
```diff
@@ -0,0 +1,12 @@
+team.1.color = red
+team.1 = red
+team.1.pos = (0,-1,103)
+team.2.pos = (0,0,-104)
+team.2.color = blue
+author = Shalen
+rotation = z
+h = 80
+team.2 = blue
+name = Four Paths
```
ClobberXD: Do you probably mean `Four Tunnels` instead? It's "tunnels" everywhere else (PR name, file names, etc.)

ClobberXD:
Map looks pretty good and error-free. A very interesting concept, I must say. The only tiny issue is the scarcity of chests. Custom chest-zones can be defined to fix this. See here for an example:

https://github.com/MT-CTF/maps/blob/9e2581925aed31e5572507a0aba8120c18f41fa0/07_serpents_pass.conf#L13-17

Note that the `from` and `to` fields are positions relative to the centre of the schem, and not absolute world coords.

ClobberXD:
> But won't the center in singleplayer world differ from that of the server?

The coords are relative to the _centre of the schem_, which means that the node at the centre of the schem is taken as `(0,0,0)`, irrespective of where the schem is placed with relation to the world.

ClobberXD:
@chkno Would you be able to attend to the requested changes? No hurries, but please update us about whether you're interested in continuing with this.

PR #8, opened 2018-12-31
Add Swamplands

xenonca:
A map where you can dig to your opponent, but as well use the bridge provided.

swamplands_screenshot1
swamplands_screenshot2
swamplands_screenshot3


ClobberXD:
A flawless map meta file! You've already got half an approval! :)

ClobberXD:
Very nice map. I've always wanted a jungle map filled to the brim with trees. The only issue is the lack of chests, as the terrain is extremely irregular. To fix this, you can define custom chest zones. See https://github.com/MT-CTF/maps/blob/9e2581925aed31e5572507a0aba8120c18f41fa0/07_serpents_pass.conf#L13-17 for an example.

ClobberXD:
Great work on the custom chest zones. There's enough chests for everyone now. :+1: 

The bases are pretty dark, though. Would you be able to place a few torches in both the bases to light them up?

xenonca:
Hm yes i can, though you have 10 torches as initial items anyway which i thought could be used for that.

ClobberXD:
The 10 torches would be required to light up the rest of the map when it gets dark. :)

MinetestSam:
The map doesn't have any problems and is good to go.
Although the red side has some apples on their side but I assume that apples will be gone so it will balance it all out

xenonca:
Wohoo! :)

PR #10, opened 2019-01-15
Add plain_battle by -sniper-

ClobberXD:
Map: Plain battle
Author: `-sniper-`
https://forum.minetest.net/viewtopic.php?p=340882#p340882

### Screenshots
screenshot
screenshot2
screenshot3


MinetestSam:
Its a good variation and it can be considered a possibility since there is no flat map

Thomas--S:
I think this might be worth a try.

If players dislike this map or spawnkilling problems occur, this map can still be removed again.

ClobberXD:
There is another flat map, but it's incomplete and the schem is cut off: #6 

I also think we should try out this map, and only then decide whether it requires any modifications.

PR #18, opened 2019-02-08
Add "Tunnel" map by -sniper-

JostP:
I made new map, it is called Tunnel. It is like caverns map, but it doesn't have caves (air in caves was replaced with stone).
There is 7x7x4 hole around the flags and beginnings of the tunnels are dug 8 nodes far (in the tunnels are 14 mese and 14 diamonds in each team). Base is made with indestructible stone brick (you can't dig them), only way out is 3x2 hole (tunnel). There is a lot of diamond and mese ore.
You spawn with mese pick, 5 stick and 20 torches. There is 15 chests in each base.
Post on forum:https://forum.minetest.net/viewtopic.php?f=12&t=19729&sid=76ec2e47a78f9f9b5573ab5a30b63c50&start=100#p342355
Here are some screenshots:
tunnel1
tunnel2
update-chests


JostP:
update: I added "chest room". 20 chests spawn in each room. There is one chest room in each base.
chests1
chests2

Thomas--S:
Based on the screenshots, I'd suggest to increase the size of the chest rooms significantly.

JostP:
**Update**: bigger chest room, 30 chests spawn in each room. There is one chest room in each base and it is made with indestructible stone brick. Room is 15x4x25 big.

tunnel-chest_room1
tunnel-chest_room2

JostP:
Is "chest room" now enough big?

ClobberXD:
Chest room is ok now. :ok_hand: 

Would it be a good idea to provide some pre-dug tunnels? Also, Should the corridors at the base entrances be longer? Otherwise, players can tunnel from anywhere, grab the flag, and run back into the tunnel. If the corridor is longer, they can only tunnel upto that much, and it'll be tougher for them to grab the flag without being noticed.

JostP:
> Chest room is ok now. ok_hand
> 
> Would it be a good idea to provide some pre-dug tunnels? Also, Should the corridors at the base entrances be longer? Otherwise, players can tunnel from anywhere, grab the flag, and run back into the tunnel. If the corridor is longer, they can only tunnel upto that much, and it'll be tougher for them to grab the flag without being noticed.

Do you think longer indestructible corridor?

ClobberXD:
Yes, it should be much longer than the current one. If not, we'd have a very tough time looking for enemies who come to get the flag.

JostP:
**UPDATE**: longer indestructible corridor, removed mese and diamonds from begining of the tunnel.

zaslonska slika 2019-02-17 12-57-30
zaslonska slika 2019-02-17 12-57-43

JostP:
If players place 3 team doors in the base entrance, enemy players cannot enter the base (if they do not dig doors).

MinetestSam:
That can a bit too powerful depending on the other team and since you get mese picks, people can easily bombard the flag taker and kill them since there is only one way to enter the base

JostP:
Is better to get bronze pick when you spawn? You need fast-dig pick because you need to make tunnels and if you want to take flag you need to barricade the indestructible tunnel with gravel or wood to prevent enemys catching you and then dig enemy's doors as fast as you can. That's why I chose mese pick.
Is mese pick better than steel sword?

ClobberXD:
Mese pick is fine by me.

ClobberXD:
Re-reviewed. Two things:

- If you're making all the base walls indestructible, you might as well make the lights indestructible. The indestructible mese lamp (`ctf_map:meselamp`) might be a great choice here, as it would adhere to the military bunker theme quite well.
- It looks like the air was replaced by stone _after_ placing the barriers. So when the barrier is removed, the parts of the barrier exposed to what was originally air nodes are now hollow, resulting in 1m-wide ugly cavities. This isn't much of a deal, and could be left as is if you don't feel like fixing it - I'm just putting it here in case someone else has a different opinion about these cavities.

JostP:
> Re-reviewed. Two things:
> 
>     * If you're making all the base walls indestructible, you might as well make the lights indestructible. The indestructible mese lamp (`ctf_map:meselamp`) might be a great choice here, as it would adhere to the military bunker theme quite well.
> 

Yes that's a good idea, I will do it
>     * It looks like the air was replaced by stone _after_ placing the barriers. So when the barrier is removed, the parts of the barrier exposed to what was originally air nodes are now hollow, resulting in 1m-wide ugly cavities. This isn't much of a deal, and could be left as is if you don't feel like fixing it - I'm just putting it here in case someone else has a different opinion about these cavities.

I will replace some of the biggest cavities and cavities near middle (tunnels could be there).

JostP:
I replaced all cavities with `//replace ctf_map:ind_glass_red default:stone` and `ind_stone with ctf_map:stone`

ClobberXD:
That's great. One more trivial nit:

screenshot_20190510_164113

The mese lamps are asymmetrical with respect to the side corridors - increasing the width of the corridor by 1m would fix this. Again, not necessarily a merge blocker, but it'd be nice to see this fixed. ;)

Thomas--S:
I reviewed this and I think that the completely missing barrier might not be good for user experience.

Why note making a completely solid barrier out of `ctf_map:ind_stone_red`?
As far as I know, this will be replaced with stone as soon as the match starts.

Otherwise, the map seems fine. :+1:

JostP:
When match starts barrier `(ctf_map:ind_stone_red)` is replaced with air, so red barrier mustn't be there.


Zaslonska slika 2019-05-10 15-23-46
Zaslonska slika 2019-05-10 15-24-26
Zaslonska slika 2019-05-10 15-25-40

JostP:
> I reviewed this and I think that the completely missing barrier might not be good for user experience.
> 
> Why note making a completely solid barrier out of `ctf_map:ind_stone_red`?
> As far as I know, this will be replaced with stone as soon as the match starts.
> 
> Otherwise, the map seems fine. +1

That is also a good idea, I didn't know what is `ctf_map:ind_stone_red` before. Map is finished now, I hope.

Thomas--S:
:+1:

JostP:
Thank you for review :)

Thomas--S:
2adea812a3b15088e81446980dc62cfb823d663c

PR #20, opened 2019-02-08
Add "Battle Island" map by DELTA_FORCE

DELTA-FORCE-CTF:
Adds Battle Island by DELTA_FORCE. A simple map, with small natural challenges for players to overcome.


MinetestSam:
The map is broken and the barriers haven't been placed correctly. And the flag positions are all messed up.
Did you manually place the barriers? 👎 
screenshot_20190209_195002

DELTA-FORCE-CTF:
Sometimes I had to do //mtschemplace a few times to get it to load properly. I will probably not place the flags, because I can't figure that out, and I don't think the mods/admin will have too hard a time with that

archfan7411:
> I don't think the mods/admin will have too hard a time with that

Your map will probably get merged a lot faster if it's ready to go ;)
I think you stated (on Discord) a difficulty getting `ctf_map` to work in map maker mode? Be sure to check that you're using it in MTG and have WorldEdit enabled. Also a good idea to clone it again in case you modified it.

DELTA-FORCE-CTF:
Yes, I have no problems with the map on my end. I just rendered it all and it looks fine

DELTA-FORCE-CTF:
Ok, I made changes. I have the map file but an incomplete .conf file, since I have no clue what it does and I don't want to mess anything up.

Thomas--S:
Please provide the .conf file generated by the ctf_map mod.

JostP:
There are missing keys in `.conf` file.

ClobberXD:
If you're going to take the manual route, you better do it properly. There are a ton of issues with the map. Can re-open once fixed.

PR #21, opened 2019-02-09
Add "The Wall" map

Thomas--S:
**Screenshots**

wall_01
wall_02
wall_03
wall_04
wall_05
Behind the two wood blocks in the last picture is the treasure chamber, where ores can be found.

The wall is indestructible, so skybridging and tunneling doesn't work in this map.


MinetestSam:
Maybe add that as hint of where the ores are? Else everything looks good 👍

Thomas--S:
How about "The treasure chamber is in the wall"?

MinetestSam:
How about "Look for a treasure room/chamber inside the wall"?

Thomas--S:
Hint added. Thanks for your suggestions :+1:

MinetestSam:
Since the map focuses on mining the ores in the center and work around that could the initial item be changed from wooden pickaxe to a stone pick? Otherwise the map is great and gets my approval 👍

Thomas--S:
I'm fine with changing this, but I don't think this is really necessary because during build time the players have enought time to find steel pickaxes from the chests and craft stone pickaxes.

The ores can only be accessed during fight time anyway.

ClobberXD:
@Gael-de-Sailly, mind giving another approval for this? :)

BrianGaucher:
When importing the schematic into a world, I had two flags stacked. Will this be fixed when in a server setting?

BrianGaucher:
My review.
Well done, there are many maps made from mapgens. We need more hand-made ones.
Nice houses for the bases. The lookout towers are cool, but i'm afraid the enemy team will destroy them quickly. I personally think there shouldn't be a passage between the two treasure rooms.

Some ideas. Make hallways in the wall, then put the stairs at the end to force players to walk down the hallway. I also think it's possible to give nodes a probability of being placed: Could this possibly be used to make the passage between rooms spawn some rounds but not others?

Thomas--S:
When I tested this map, no stacked flags occurred.

At first, I also thought about blocking the passage between the treasure rooms, but then I realized that this isn't necessary because there are more than enough ores in each treasure room, so every team can get enough. Secondly, I think it is a little bit more challenging if there is the enemy team nearby. Thridly, the passage is blocked with destroyable wood anyway, so the connection between the rooms won't be there immediately.

I didn't really understand your idea regarding hallways, sorry.

MinetestSam:
> When importing the schematic into a world, I had two flags stacked. Will this be fixed when in a server setting?

Are you placing the schem in a singlenode world or running it on the CTF subgame? The map works fine from my testing and there are no stacked flags

BrianGaucher:
Singlenode, but all future testing should be one server. ANAND told me how.

On February 25, 2019 2:07:27 AM EST, MinetestSam <notifications@github.com> wrote:
>> When importing the schematic into a world, I had two flags stacked.
>Will this be fixed when in a server setting?
>
>Are you placing the schem in a singlenode world or running it on the
>CTF subgame? The map works fine from my testing and there are no
>stacked flags
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/MT-CTF/maps/pull/21#issuecomment-466895530

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Thomas--S:
Thanks for merging!

PR #22, opened 2019-02-09
Add "Karsthafen town" by solars, Peters and -sniper-

JostP:
Second part of karsthafen. Its name is **Karsthafen town centre**. Buildings are destructible. There is church on the map. It is indestructible and it doesn't have any entrance, so players won't fight in it. I couldn't demolished church, because I don't think it is moral. Also I think that author of the Karsthafen won't agree with removing the church. I haven't seen a muslim IRL who complained when he saw a church.

I suggest to test this part of the map and if there will be problems remove it from the game.

karsthafen-town_centre2
karsthafen-town_centre6
karsthafen-town_centre7


Thomas--S:
I'm strongly opposed to adding the church for the reasons given in #15.

Although the measures taken to prevent problems might help to some degree, they won't be sufficient to prohibit problems.

I'd approve the solution originally proposed in #15, or a completely empty space, or the church being replaced by another building in a fitting style (e.g. a small castle).

JostP:
There is a lot of crosses and also a graveyard in this part of the map. I don't want to completely change the map and solars' beautiful artwork. If you think that this map is not suitable for the game then we should find someting else.

ClobberXD:
Would it make sense to rename this to "Karsthafen town". The one with the lake can be renamed  "Karsthafen lake" (or maybe "Karsthafen bridge"?).

Thomas--S:
@JostP I agree with you that it is probably better to find a less controversial part of the Karsthafen map, maybe from another town there. It's true that the graveyard might also lead to problems.

MinetestSam:
The church is truly a controversial building as it can cause conflict. Some may destroy the church and some may not and that will spark controversies related to the church so just to be on the safer side either replacing the church with a building or removing it completely will be better.

JostP:
I found "Red city" in solars' Karsthafen map. It is oriental town, so there is mosque. If you want, I can make map from this part and we will have map with church and map with mosque. Both religions will be satisfied.
As I said, i don't want to change map and I wish to **test map first** and see if there will be problems.


zaslonska slika 2019-02-10 16-48-41

Thomas--S:
> Both religions will be satisfied

I think it will lead to even more problems because now there will also be anti-muslim problems.

> **test map first** and see if there will be problems

But this is to late then because there were already problems.



As an alternative, I suggest to take for example a part of the Red city without religious buildings and use this as a map.

JostP:
@Thomas--S : what problems there were already?

Thomas--S:
I think there was a missunderstanding: I didn't mean that there were already problems **now**, but when the map leads to problems.

archfan7411:
Agreed with Thomas-S, the issue is not so much the religious players being unsatisfied with representation, but players that would willingly disrespect any religion-themed places in a map, leading to controversy with the religious players. The church and mosque in the sections are both beautiful buildings, but not suited for a game of this nature, IMO. Replacing it with a park-like environment, complete with paths and trees, is what I'd suggest.

JostP:
As I said, I don't want to change artwork or remove church, because both isn't moral for me.

If you don't want to test it, then we should add only `Karsthafen bridge` (#23 ) first.

BTW: Players often build a cross in game and nobody worry about it or destroy it
cross
.

JostP:
### **UPDATE: DIFFERENT PART OF THE ORIGINAL CITY.**

karsthafen_town1
zaslonska slika 2019-02-14 14-49-45
zaslonska slika 2019-02-14 14-49-58

karsthafen_town.conf
```diff
@@ -0,0 +1,21 @@
...
+r = 115
+h = 140
+rotation = x
+team.1 = red
+team.1.color = red
+team.1.pos = (46,-4,105)
+team.2 = blue
+team.2.color = blue
+team.2.pos = (-30,-5,-83)
+initial_stuff = default:pick_wood,default:sword_stone,default:torch 3
```
Thomas--S:
```suggestion
initial_stuff = default:pick_stone,default:sword_stone,default:torch 3
```

xenonca:
If it gets abused it would probably be good to have indestructible nodes around the lava in the light house near blue's base, since its pretty high up there and could even flow very close the base.

xenonca:
For clarification:

kth

JostP:
Should I remove wood doors too?

xenonca:
This is probably just an aesthetic thing but there are some fence pieces missing it seems:

kht

Thomas--S:
I don't think that the doors must be removed necessarily in this map.

If players are too confused, they can be removed later, too.

JostP:
I had removed doors before I saw your comment. So in this update there are removed doors and indestructible light house.

ClobberXD:
Is this map still WIP? Has the lava been sealed with ind. nodes?

JostP:
Yes

JostP:
I mean "yes" for ind. nodes

ClobberXD:
@Thomas--S' approval still holds, as he confirmed on Discord.

PR #23, opened 2019-02-10
Add "Karsthafen bridge" by solars, Peters and -sniper-

JostP:
First part of the Karsthafen. It has bridge and 2 lakes. Bridge separates 2 parts of the city.
This is updated version of #15 .

karsthafen_bridge1
karsthafen_bridge2
karsthafen_bridge5


JostP:
> The license field in the `.conf` requested in the [forum post](https://forum.minetest.net/viewtopic.php?f=12&t=6642&start=275#p341671) is missing.

Is ok if I put this: `license = CC BY SA 4.0` into `.conf` file?

karsthafen_bridge.conf
```diff
@@ -0,0 +1,19 @@
+name = Karsthafen bridge
+author = solars, Peters and -sniper-
+license = CC BY SA 4.0
```
ClobberXD:
```suggestion
license = CC-BY-SA 4.0
```

Thomas--S:
Actually, I'd still prefer the line suggested by solars:
`author = solars, Peters and -sniper- from the minetest.net forum`
Otherwise, it isn't clear who solars is, as this is only a nickname.


> and state if the work has been modified or not

Where is this information in the .conf file?

Please note: I am not a lawyer.

JostP:
> Actually, I'd still prefer the line suggested by solars:
> `author = solars, Peters and -sniper- from the minetest.net forum`
> Otherwise, it isn't clear who solars is, as this is only a nickname.

Yes, I agree with you, I will edit `author` key.

> > and state if the work has been modified or not
> 
> Where is this information in the .conf file?

I don't think if this is needed, because I didn't modity anything, I only took part of the original map.

ClobberXD:
> Otherwise, it isn't clear who solars is, as this is only a nickname.

Good point. I also suggest modifying the line to read:

`author = solars, Peters and -sniper- from forum.minetest.net`

> > Where is this information in the .conf file?
>
> I don't think if this is needed, because I didn't modity anything, I only took part of the original map.

This is what I originally thought, but I'm having second thoughts. e.g. do the barriers count as modifications to the original map? They are temporary, and disappear during the game.

JostP:
> > Otherwise, it isn't clear who solars is, as this is only a nickname.
> 
> Good point. I also suggest modifying the line to read:
> 
> `author = solars, Peters and -sniper- from forum.minetest.net`
> 

Ok, I will also edit `author` key in #24 like this.
> > > Where is this information in the .conf file?
> > 
> > 
> > I don't think if this is needed, because I didn't modity anything, I only took part of the original map.
> 
> This is what I originally thought, but I'm having second thoughts. e.g. do the barriers count as modifications to the original map? They are temporary, and disappear during the game.

I don't think that the barriers count as modifications, also solars didn't said anything about it.

JostP:
Now when we have only one part of Karsthafen, should we give it name "Karsthafen" (`name = Karsthafen`) instead of "Karsthafen bridge"?

Thomas--S:
I suggest to stay with the name Karsthafen bridge so we can avoid confusion with the name of the whole map and we also have the opportunity to easily add other parts later.

Regarding the modifications: In my opinion – and again: I am not a lawyer – replacing bookshelves with wood is a modification.

I suggest to add a `note` or `comment` field to the `.conf` file, where we can write that changes were made.

JostP:
> Regarding the modifications: In my opinion – and again: I am not a lawyer – replacing bookshelves with wood is a modification.
> 
> I suggest to add a `note` or `comment` field to the `.conf` file, where we can write that changes were made.

Is possible to add comment into `.conf` file? Sorry i forgot that I:
- removed rail
- replaced xpanes and obsidian_glass with glass
- removed mese_post_light
- replaced bookshelf with wood
- replaced metal blocks (gold block, steel block ...) with wool
- removed grass
- removed wood signs
- replaced obsidian with black wool
- removed wooden door
- removed diamond ore (near 500 nodes)
- removed mese ore and mese blocks.

Image what I removed/replaced:
zaslonska slika 2019-02-11 16-51-01

JostP:
Add this into `.conf` file?
`modifications = replaced/removed nodes which are not supported in CTF, placed barriers`

Thomas--S:
Maybe you should also add that the map was cropped and only a small part is used.

JostP:
`modifications = replaced/removed nodes which are not supported in CTF, placed barriers. Only small part of the original map is used.`
Ok ?

Thomas--S:
:+1: Looks good to me. I'd only suggest to capitalize the first letter :)

Thomas--S:
I reviewed this map again, and in general, I approve this map. There is only one issue left:

The church and especially the altar are still accessible near red base. Would it be possible, to decrease the size of the map a little bit, so that the church is outside of the barrier?

JostP:
Is this really needed? It takes me near half hour to do this (making map, replacing not-supported nodes, testing)?

BTW did you see my update in #22 ?

Thomas--S:
#22 is already approved :)

Sorry for causing inconvenience, but I think the church might lead to problems. If only the rear part of the church was accessible or the church was of no tactical value, maybe I'd think differently.

But in current state, the altar is accessible and the church is of tactical value because it can be used to shoot at the base.


One final thought regarding the `modifications` field: Do you think we should not only mention the barrier, but also include the flag and chests?

JostP:
> One final thought regarding the `modifications` field: Do you think we should not only mention the barrier, but also include the flag and chests?

No, chests are part of CTF game and they aren't part of the map. Chests and flags appear when you start map in CTF game.

JostP:
**Update: smaller map**

JostP:
Do ANAND need to make command `/map` before merging?

Thomas--S:
> Do ANAND need to make command `/map` before merging?

I think this might be the best solution.

Thomas--S:
Thank you very much for updating the map again! :+1:

JostP:
If you merge all my 3 city maps, they will be in game one after the other, so there will be 3 city maps in a row. This is quite pointless IMO.
Also `Desert spikes` and `Ice age` are in the row and they are similar to each other(because you spawn with steel sword). Also `Coast of Nostalgia` and `Plains of the past` are similar to each other and they are in a row.

I suggest to make sequence of maps if possible so there will be different maps one after the other.

JostP:
Now when you approved 2 parts of Karsthafen, I can make one part indestructible if you want. Now when indestructible stairs and wool works, I know how to do that. Also I an not as busy as ANAND is.

Thomas--S:
I'd actually prefer it if we leave both maps destructible and see how players behave.
If necessary, they can be made indestructible later, too.

ClobberXD:
-sniper- is the sole author of all the Karsthafen maps. The maps were adapted from another world which was created by solars and Peters, but that doesn't make them the authors of the CTF map, as in they didn't physically (re)build all this with -sniper-. It's better if they're attributed in the `license` field instead.

Standard disclaimer: I'm not a lawyer, but what I said probably makes sense from a legal point of view :3

Thomas--S:
Will re-review soon.

JostP:
> -sniper- is the sole author of all the Karsthafen maps. The maps were adapted from another world which was created by solars and Peters, but that doesn't make them the authors of _this_ CTF map, as in they didn't physically (re)build all this with -sniper-. It's better if they're attributed in the `license` field instead.
> 
> Standard disclaimer: I'm not a lawyer, but what I said probably makes sense from a legal point of view :3

Solars and Peters have been building Karsthafen map since 2013 and they spent more than hundred hours (I guess) to build whole city. I spent only 2 hours to make playable CTF map. So I think, it is fairly to attribute them on the most visible place - `author` key. Players will see them always when match on this map starts. 

If we want players to know who is an author of CTF map and who are authors of the original map, then we should modify `author` key like this: `author = -sniper- (CTF map maker) and solars, Peters from forum.minetest.net (original map makers)`. This is my opinion.

JostP:
Should I add anything new what requests map catalog? Maybe I should modify license key: `license = CC-BY-SA 4.0. This CTF map is addapted from Karsthafen map, created by solars and Peters. You can find it on forum.minetest.net.`

Thomas--S:
I suggest the following for the `.conf` file (analog for the other two maps):

```
author = -sniper- (CTF map maker) and solars and Peters from forum.minetest.net (original map makers)
license = CC-BY-SA 4.0. This CTF map is adapted from Karsthafen map, created by solars and Peters. You can find it on forum.minetest.net.
others = This map is a part of the town "Karsthafen" from the map "Karsthafen" by solars from forum.minetest.net
```

Please note: I am not a lawyer.

@ClobberXD, @rubenwardy: What do you think about this suggestion?

ClobberXD:
- The `author` field corresponds to the creator of the CTF map. If the map is a derivative of an existing work, the creators/authors of the original work should be attributed in `license`. So unless solars and Peters helped -sniper- with the actual schem, they can't be credited as authors.

- `others` seems to contain the exact same stuff contained in `license`, apart from `CC-BY-SA 4.0`, which seems unnecessary to me. Also `license` is the field to be used for giving attribution to the original work, and hence this field can be removed.

- The attribution seems a little too verbose, and can be made briefer. (See below)

****

Quoting from Creative Commons' official wiki page on [Best practices for attribution](https://wiki.creativecommons.org/wiki/best_practices_for_attribution): 

> A good rule of thumb is to use the acronym TASL, which stands for Title, Author, Source, License.
> 
> Title - What is the name of the material?
> 
>     If a title was provided for the material, include it. Sometimes a title is not provided; in that case, don't worry about it.
> 
> Author - Who owns the material?
> 
>     Name the author or authors of the material in question. Sometimes, the licensor may want you to give credit to some other entity, like a company or pseudonym. In rare cases, the licensor may not want to be attributed at all. In all of these cases, just do what they request.
> 
> Source - Where can I find it?
> 
>     Since you somehow accessed the material, you know where to find it. Provide the source of the material so others can, too. Since we live in the age of the Internet, this is usually a URL or hyperlink where the material resides.
> 
> License - How can I use it?
> 
>     You are obviously using the material for free thanks to the CC license, so make note of it. Don't just say the material is Creative Commons, because that says nothing about how the material can actually be used. Remember that there are six different CC licenses; which one is the material under? Name and provide a link to it, eg. http://creativecommons.org/licenses/by/4.0/ for CC BY.
> 
>         → If the licensor included a license notice with more information, include that as well.

See also:
- [Good attribution for slightly modified material](https://wiki.creativecommons.org/wiki/best_practices_for_attribution#This_is_a_good_attribution_for_material_you_modified_slightly)
- [Good attribution for a derivative work](https://wiki.creativecommons.org/wiki/best_practices_for_attribution#This_is_a_good_attribution_for_material_from_which_you_created_a_derivative_work)
- [Title, Author, Source, License](https://wiki.creativecommons.org/wiki/best_practices_for_attribution#Title.2C_Author.2C_Source.2C_License)

Thomas--S:
I think that solars should be attributed in `author`, too, as they are the one who built this map. -sniper- "only" converted it.

The text in the `others` field was mainly meant to indicate which part of the map was used, so that interested players can find it more easily. But I agree with you that the wording can be improved.

How about the following?

```
author = -sniper- (CTF map maker) and solars and Peters from forum.minetest.net (original map makers)
license = CC-BY-SA 4.0. This CTF map is adapted from Karsthafen map, created by solars and Peters. You can find it on forum.minetest.net.
others = The part used in this CTF map belongs to the town "Karsthafen" in the original map.
```

JostP:
What about modifications?

ClobberXD:
Here's a briefer attribution:

```
license = CC-BY-SA 4.0. Adaptation of "Karsthafen" (<link to Karsthafen's forum topic>), by solars and Peters from the Minetest forums (forum.minetest.net)
```

Note that I've omitted phrases like "This CTF map has been...", "...You can find it...", etc. as the attribution should be brief, and to the point.

> What about modifications?

Not needed, as CC-BY-SA 4.0 does not require the modifications made to the original work to be listed out.

Thomas--S:
So can we agree on the following configuration?

```
author = -sniper- (CTF map maker) and solars and Peters from forum.minetest.net (original map makers)
license = CC-BY-SA 4.0. Adaptation of "Karsthafen" (https://forum.minetest.net/viewtopic.php?f=12&t=6642) by solars and Peters, from the Minetest forums (forum.minetest.net)
others = The part used in this CTF map belongs to the town "Karsthafen" in the original map.
```

ClobberXD:
I'm still not happy with `author` being so verbose. Could the extra info be moved to `others`, leaving `author` with `-sniper-, solars, Peters` which is less wordy. We'll end up with the following result:

```conf
author = -sniper-, solars, and Peters
license = CC-BY-SA 4.0. Adaptation of "Karsthafen" (https://forum.minetest.net/viewtopic.php?f=12&t=6642) by solars and Peters, from the Minetest forums (forum.minetest.net)
others = This map was taken from Karsthafen town region in the original Karsthafen map (<forum topic URL>) by solars and Peters, and converted into a CTF map by -sniper-.
```

Thomas--S:
:+1: for the configuration suggested by ANAND.

However, I think that I think that the forum url doesn't have to be repeated in the others field.

karsthafen_bridge.conf
```diff
@@ -0,0 +1,21 @@
+name = Karsthafen bridge
+author = solars, Peters and -sniper- from forum.minetest.net
```
Thomas--S:
```suggestion
author = -sniper-, solars, and Peters
```

karsthafen_bridge.conf
```diff
@@ -0,0 +1,21 @@
...
+license = CC-BY-SA 4.0
```
Thomas--S:
```suggestion
license = CC-BY-SA 4.0. Adaptation of "Karsthafen" (https://forum.minetest.net/viewtopic.php?f=12&t=6642) by solars and Peters, from the Minetest forums (forum.minetest.net)
```

karsthafen_bridge.conf
```diff
@@ -0,0 +1,21 @@
...
+modifications = Replaced/removed nodes which are not supported in CTF, placed barriers. Only small part of the original map is used.
```
Thomas--S:
```suggestion
others = This map was taken from Karsthafen town region in the original Karsthafen map (<forum topic URL>) by solars and Peters, and converted into a CTF map by -sniper-.
```

JostP: (<forum topic URL>)?

Thomas--S: Sorry, I actually wanted to delete this before submitting the comment.

karsthafen_bridge.conf
```diff
@@ -0,0 +1,21 @@
...
+r = 100
+h = 130
+rotation = x
+team.1 = red
+team.1.color = red
+team.1.pos = (-27,1,96)
+team.2 = blue
+team.2.color = blue
+team.2.pos = (39,-1,-96)
+initial_stuff = default:pick_wood,default:sword_stone,default:torch 3
```
Thomas--S:
```suggestion
initial_stuff = default:pick_stone,default:sword_stone,default:torch 3
```

PR #24, opened 2019-02-11
Add "Hunting season" by solars, Peters and -sniper-

JostP:
Hello
I made new map. I made it from part of solars' Karsthafen map. Its name is **hunting season**. It is small village  made of wood and red stone. 

Screenshots:

zaslonska slika 2019-02-11 17-40-47
zaslonska slika 2019-02-11 17-41-33
zaslonska slika 2019-02-11 17-41-02
zaslonska slika 2019-02-11 17-40-23


Thomas--S:
This village is called "Jägerhof" in the original map, isn't it?

hunting_season.conf
```diff
@@ -0,0 +1,13 @@
+name = Hunting season
+author = solars, Peters and -sniper-
```
Thomas--S: Actually, solars asked for this line: (https://forum.minetest.net/viewtopic.php?f=12&t=6642&start=275#p341671)

```suggestion
author = solars, Peters and -sniper- from the minetest.net forum
```

JostP:
What do you think about changing `author` and adding `hint` like this:
`author = -sniper-`
`hint = Map is adapted from a Minetest world made by solars and Peters. You can find it on forum.minetest.net. Map is distributed under the CC-BY-SA 4.0 license`

Tell me what do you think. Is this correct in english?

JostP:
> Jägerhof

What does "Jägerhof" mean?

Thomas--S:
> What do you think about changing `author` and adding `hint` like this:
`author = -sniper-`
`hint = Map is adapted from a Minetest world made by solars and Peters. You can find it on forum.minetest.net. Map is distributed under the CC-BY-SA 4.0 license`
>
>Tell me what do you think.

I think this is a good idea, but ultimately it's solars's decision.

---

Jägerhof means something like Hunters' Yard. Of course, I prefer the English name provided by you (Hunting Season) over the original name.

---

I already reviewed this map and I think it looks very promising.
As soon as the `.conf` file is fixed according to solars's wishes, I'll approve this map.

JostP:
> As soon as the `.conf` file is fixed according to solars's wishesAs soon as the `.conf` file is fixed according to solars's wishes

https://creativecommons.org/licenses/by-sa/4.0/deed.en :
Attribution — You must give appropriate credit, provide a link to the license, and indicate if changes were made. You may do so in any reasonable manner, but not in any way that suggests the licensor endorses you or your use. 

I am author of CTF map and solars and Peters are authors of the original map.
I can ask solars if you want.

Thomas--S:
I think it's best to ask solars.

JostP:
**update** Add hint and change "author" key

Do you have any suggestion about changing hint? Please check if hint is written correct in english.

ClobberXD:
I'm not happy with such a large hint. A hint is meant to provide players with a tip on how the map is meant to be played. The `author` field is logically more suitable for this, but even that isn't the best choice - keep in mind that both the author field and the hint field will be displayed in game. I prefer to not have large fields as these flood the chat at the start of every match. It also looks like the attribution should link to the license itself, which makes it even more complicated.

At the moment I can think of having a `/maps` chat-command that displays a formspec that lists all maps, with name, author, license, etc. We could then add a description field to the conf file that will be displayed in this formspec.

I could work on this, but the map can't be merged until then.

ClobberXD:
Accidentally dismissed approval of @Thomas--S. Apologies.

MinetestSam:
Why does numbering matter btw? Does it decide the sequence?

ClobberXD:
> Why does numbering matter btw? Does it decide the sequence?

Yes, see map parsing function [here](https://github.com/MT-CTF/capturetheflag/blob/5daefadc7e8e3a0982fa012fd7544f43fa4ecba7/mods/ctf/ctf_map/schem_map.lua#L46) - prefixing by numbers ensures that the maps are registered in the same order.

JostP:
> > Why does numbering matter btw? Does it decide the sequence?
> 
> Yes, see map parsing function [here](https://github.com/MT-CTF/capturetheflag/blob/5daefadc7e8e3a0982fa012fd7544f43fa4ecba7/mods/ctf/ctf_map/schem_map.lua#L46) - prefixing by numbers ensures that the maps are registered in the same order.

So if I understand, `forest map` will be after `plain battle` and `river valley` after `forest map` ?

ClobberXD:
> So if I understand, `forest map` will be before `plain battle` and `river valley` before `forest map` ?

Not really, the order of registration is the order in which the maps were merged, not the PRs created. So ideally, the number to be prefixed can only be decided when merging.

AFAIK, maps would still work without prefixing the filenames with numbers, but they just won't be consistent with respect to order of registration.

BrianGaucher:
> At the moment I can think of having a `/maps` chat-command that displays a formspec that lists all maps, with name, author, license, etc. We could then add a description field to the conf file that will be displayed in this formspec.

I agree with this. It would allow for longer descriptions, avoid licensing issues, provide useful hints, and would be good to let players see all the maps.

ClobberXD:
I'm not quite sure how the `modifications` field would help. We could keep it until the /maps feature has been merged, but then players won't be able to see/access the license from within the game, and that'd be a violation of the license.

This map, and the other ones from Karsthafen will have to be put on hold.

hunting_season.conf
```diff
@@ -0,0 +1,15 @@
...
+r = 115
+h = 140
+rotation = z
+team.1 = red
+team.1.color = red
+team.1.pos = (-14,-11,105)
+team.2 = blue
+team.2.color = blue
+team.2.pos = (17,-2,-97)
+initial_stuff = default:pick_wood,default:sword_stone,default:torch 3
```
Thomas--S:
```suggestion
initial_stuff = default:pick_stone,default:sword_stone,default:torch 3
```

Thomas--S:
94e7fbfe0283036449e0275c0b6c3bf4c2b65ce3

PR #26, opened 2019-03-12
Add "Abandoned Isles" map by gimp

mrowlbio:
I made the paths indestructible.

I will deal with the custom chest zones later, The map is done, can you add it so we can get feedback on how it plays, then I will do the chest zones.


Thomas--S:
Previous comments are in #3

xenonca:
Tested it, seems a nice new map to play!

JostP:
I can make chest zones for you if you want. It is not so hard and it would be good if there were enough chests. I am away this week, I can make it next week.

ClobberXD:
Sorry for the late review. Here are my thoughts:

1) Map looks pretty good overall.
2) There's a severe lack of chests as already noted.
3) Players are given 50 sandstones by default, but they aren't provided with tool(s) to break them?

I can cope with 2, as long as 3 is fixed - i.e. a steel pick should be provided by default. I'd still prefer custom chest zones in the long-term though.

ClobberXD:
@mrowlbio You around? Are you still interested in taking up this PR?

ClobberXD:
Continued in #52

PR #30, opened 2019-04-09
update plain_battle

JostP:
I made update for my plain battle map:
-more useful buildings (wall and fortress)
-6 meses in mine in each hunter's house
Zaslonska slika 2019-04-09 17-53-55
Zaslonska slika 2019-04-09 17-52-46
Zaslonska slika 2019-04-09 17-52-54
Zaslonska slika 2019-04-09 17-55-00
Zaslonska slika 2019-04-09 17-52-19
Zaslonska slika 2019-04-09 17-51-35


Thomas--S:
I reviewed this and the update seems to be very promising.

I suggest to remove the doors and furnances from the hunter's houses. They don't work (because of missing metadata). This will confuse players.

I think that it would be easier for players if the directions were marked with red/blue wool. I think of the following places:
 - The very center castle made of stonebrick
 - The desert stonebrick castles
 - Maybe the forest watchtowers

It might also be a good idea to place the old wooden castles from the previous version at the very empty, forest-less side of the map.

JostP:
Wool mark, remove doors and furnaces 

Zaslonska slika 2019-05-11 09-39-59

Thomas--S:
I think that using breakable wool with breakable structures might be a little bit better.

Furthermore, I'd prefer it, if the wool colors were visible in the inside room of the center stonebrick castle.

Are you going to re-add the wooden castles, too, as suggested above? It isn't mandatory, I'd only like to know whether you'll do this or not, so I know when this update is ready for merge.

JostP:
> I think that using breakable wool with breakable structures might be a little bit better.
> 
I don't think this is a good idea, if players dig wool they can place it everywhere (on the enemy castle for example).

> Furthermore, I'd prefer it, if the wool colors were visible in the inside room of the center stonebrick castle.
> 

I will fix that
> Are you going to re-add the wooden castles, too, as suggested above? It isn't mandatory, I'd only like to know whether you'll do this or not, so I know when this update is ready for merge.

No, there won't be battles in the forest and I don't think we need it.

Thomas--S:
> I don't think this is a good idea, if players dig wool they can place it everywhere (on the enemy castle for example).

I don't agree with you here. Re-placing the wool can be used as a strategy, e.g. to irritate the enemy etc. It can also be used to mark additional structures.

> >  Furthermore, I'd prefer it, if the wool colors were visible in the inside room of the center stonebrick castle.
>
> I will fix that

Thanks :)

> No, there won't be battles in the forest and I don't think we need it.

Ok. I only thought the old castles were to good to be deleted :)

But I can understand your point of view, and I'm totally fine with it.

Thomas--S:
:+1: for the new wool position, but I'd still prefer the breakable one.

Thomas--S:
:+1: Very nice update! Thanks!

JostP:
Thank you for review :).

10_plain_battle.conf
```diff
@@ -8,5 +8,5 @@ rotation = z
 h = 140
 team.2 = blue
 name = Plain battle
-initial_stuff = default:pick_wood,default:sword_stone,default:torch 5
+initial_stuff = default:pick_stone,default:sword_stone,default:torch 5
```
Thomas--S: Can you please remove this change as this is already contained in master?

JostP:
I saw yesterday that gun (small black one) destroys wool. Wool is on castles at the middle, where players will shoot a lot. Maybe it's better idea to make wool indestructible?

Should I add `10_plain_battle.png` to `.conf file`?

And should I add screenshots to karsthafen maps?

ClobberXD:
Adding indestructible wool on top of a destructible castle would result in floating wool if and when the castle below is destroyed, and wouldn't look nice. If you want to make the wool indestructible, I suggest you make the castle indestructible too. Or you can just leave it as it is, either option is fine with me.

Thomas--S:
I definitely prefer the whole castles, including the wool, being destructible.
Destroying the wool with guns doesn't seem like a big problem to me.

I already prepared a screenshot for this update of Plain Battle, but feel free to include your own.

JostP:
> I definitely prefer the whole castles, including the wool, being destructible.
> Destroying the wool with guns doesn't seem like a big problem to me.
> 
> I already prepared a screenshot for this update of Plain Battle, but feel free to include your own.

I like your screenshot, I just wanted to ask if I have to add **line** `10_plain_battle.png` to `.conf `file. Thank you for your review too ClobberXD.

ClobberXD:
@JostP Yes, please do add `screenshot = 10_plain_battle.png` to the map meta.

PR #31, opened 2019-05-12
Add screenshots

Thomas--S:
Screenshot for Plain Battle already shows the updated version, that's why it isn't included in the .conf file yet.


ClobberXD:
The filenames don't comply with Minetest's texture nomenclature, but that's even better in this case, I suppose.

Thomas--S:
I chose this naming syle to be consistent with the naming of all map files.

Conflicts with Minetest texture file names are nearly impossible, I think.

ClobberXD:
Serpent's Pass needs a screenshot. It won't show up in the maps catalog, as the map is disabled, but it should have a screenshot ready, if and when enabled.

Thomas--S:
Serpent's Pass screenshot included :smiley:

ClobberXD:
The resolutions of the screenshots are 600x400. While the ratio is fine, I think higher-res images would be nice to have. I suggest a standard size of 1920x1080, which is a very popular resolution for both images and videos alike, and have an aspect ratio of 16:9 which is also the most popular aspect ratio at the moment.

Do note - this is just a suggestion, and this isn't required for approval. I'd like to know your opinion as well.

Thomas--S:
While I think that a size of 1920x1080 might be nice to look at, I think that many players would be afraid of the long download times.

I chose 600x400 because it is actually big enough for the preview image on most screens, I think.
600x400 was also the suggested size for a mod's screenshot.png, as far as I know.

ClobberXD:
That's fine with me, then.

PR #32, opened 2019-06-18
Add Ancient Pyramids

xenonca:
New map with 4 pyramids (plus a tiny one) and a small labyrinth inside the pyramids.

15_ancient_pyramids_orig
screenshot_20190617_155301
screenshot_20190617_155228


Thomas--S:
I think this map is an interesting idea.

However, I'm not really sure if it is very enjoyable to play on this map.

It seems like this map rewards tunneling; although this technique is usually not welcome.

In the current state, I'd prefer to stay neutral.

---

Maybe it would be a good idea to separate the two pyramids a bit more, so that they can be defended better. I'd also suggest to make tunneling harder by adding more unbreakable materials where appropriate. The labyrinths could be made less repetitive and confusing for the defending team. Maybe it could be used as some sort of pre-built defending structure?

JostP:
I reviewed your map:

Maybe it would be good if you remove this glass (for less lag ) because it is redundant.

Zaslonska slika 2019-06-19 13-38-02

There are 2 destructable mese and one destructable diamond on this map.

Zaslonska slika 2019-06-19 13-46-17

When I replaced destructable nodes with air with WorldEdit, hole appeared inside indestructible wall:

Zaslonska slika 2019-06-19 13-46-35

I agree with Thomas-S about tunneling, it is really silly because it usually makes games short and ends of matches uninteresting.

However, maps looks promising good and it works :) .

xenonca:
Hey cool, thanks for the feedback!
Sure, separating the pyramids might indeed be a good idea.
However tunneling is in this case a valid tactic on this map, I also think it will be different to other maps, since the bases are a bit elevated and there is already a prebuilt direct underground connection between the two bases. That's the reason why you get a shovel as initial stuff. I'd personally leave it as it is and see how it goes, still able to change that afterwards.

> Maybe it would be good if you remove this glass (for less lag ) because it is redundant.

Yea I need to figure that out how to do that properly.
Though I don't quite get the problem with the (in-/)destructible mese/dias, they're made like that on purpose. There's a lot of sand on the map so a diamond/mese shovel might seem useful. While the mese nodes at the base are decorative parts.

> When I replaced destructable nodes with air with WorldEdit, hole appeared inside indestructible wall:

Huh?

JostP:
> > When I replaced destructable nodes with air with WorldEdit, hole appeared inside indestructible wall:
> 
> Huh?

WorldEdit mod, it is very useful for fast building.
https://github.com/Uberi/Minetest-WorldEdit

xenonca:
I know WorldEdit and used it for this map, I'm just not getting how that happened/what that is that "hole"

JostP:
> I know WorldEdit and used it for this map, I'm just not getting how that happened/what that is that "hole"

That "hole" was destructable before I replaced sand with air (` //replace default:sand air`).

xenonca:
Separated the pyramids, removed the extra glass under the map and reworked the labyrinth:

ap1
ap2

ClobberXD:
This map looks pretty interesting. Here are my thoughts:

1) The bases are underground (i.e. surrounded by solid nodes on all sides) and are vulnerable to undetected enemy tunnels from all sides.
2) The pyramids and labyrinths are nice, but won't serve their purpose if they're diggable, IMO. Enemies could just dig from anywhere, or dig up the whole pyramid, exposing the base from all sides.
3) The other diagonal of the map looks empty, despite having two giant pyramids along it. Players should be encouraged to explore the whole map, instead of tunnelling straight

Points 1 and 2 are supplementary. Both can be fixed by making the base walls, pyramids, and the labyrinths indestructible. Enemies would only be able to enter via the 4 designated entrances.

Point 3 can be addressed by providing players some sort of an incentive to go to the other corners of the map. Maybe rare ore clusters? Maybe make the hollow pyramids double up as chest rooms (custom chest zones)? Maybe both?

xenonca:
> 1./2. ... are vulnerable to undetected enemy tunnels from all sides.
The floor is indestructible, imo it would be boring to make all sides including the labyrinth indestructible nodes because it would just take some matches on that map and the players would know the whole labyrinth. Leaving the nodes destructible would make it a challenge every match to check if there's a new tunnel that has been made by the enemy or not, similar to caverns. With the difference that they would have to get out of the pyramid on the surface since the floor is indestructible.

> 3. ...
Hm that's true, would it be good to change the initial_stuff to stone swords and add iron clusters to the pyramids? Because adding mese or diamonds would change the gameplay very much. However that can also be a positive gameplay change. Undecided on what to add, what would you suggest? (Please don't let me go trough adding chest zones again... X) )

ClobberXD:
> Undecided on what to add, what would you suggest? (Please don't let me go trough adding chest zones again... X) )

So mean :(

Well, then we can try to add a physical hurdle in the centre, making the players want to go around it. Maybe remove the pyramids on the sides, and add a pyramid right at the centre? Players can go over the pyramid if they want to, but most would prefer going around to remain undetected. This approach provides players a stealth bonus if they take the longer route.

xenonca:
Ok good idea too, will work on it!

17_ancient_pyramids.conf
```diff
@@ -0,0 +1,14 @@
...
+team.1.color = red
+team.1 = red
+author = Xenon
+rotation = x
+h = 40
+team.2 = blue
+name = Ancient Pyramids
+initial_stuff = default:pick_steel,default:sword_steel,default:torch 15,default:shovel_steel
+r = 86
+hint = Beware the Pyramid's Labyrinth!
```
ClobberXD:
```suggestion
hint = Beware of the Pyramids' Labyrinth!
```

JostP:
it's WIP, isn't it?

ClobberXD:
@xenonca Bumpity-bump... Still WIP?

xenonca:
Yes

ClobberXD:
Closing due to inactivity. Can be re-opened upon progress.

PR #35, opened 2019-07-07
Add Big Ocean by gameit

MinetestSam:
https://forum.minetest.net/viewtopic.php?p=350983#p350983


JostP:
I think stone under map should be removed or replaced with water because players can freely make tunnels and map lose its point.

 Zaslonska slika 2019-07-08 10-32-21

Why steel pick? There is only a little stone on surface.

MinetestSam:
I think it would be better to write the review on the forum page since the user is not on github

MinetestSam:
gameit's response to -sniper-

```
I can remove the stone, but the steel pick-axe isn't important that way.

Under a real ocean are much things like coal and iron. If I make ores in the stone too, the map would be more realistic and you can use the steel pick-axe to mine.

What is the better way??
```

MinetestSam:
I won't be able to regularly update this PR, so I would like someone else to open this PR (I am really sorry).

ClobberXD:
PR needs adoption. Re-opening for attention.

JostP:
continue #48

PR #37, opened 2019-07-12
Improve The Wall Map

Thomas--S:
Add custom chest zones to provide more chests
Add steel pick to start items
Make the mese above the doors breakable as originally intended


ClobberXD:
Nice

Are these many chests meant to spawn? If this is what you intended, then that's fine.

screenshot_20190713_070236

(I've removed the trees to make all the chests visible)

Thomas--S:
I think that `n = 50` is a bit too less because the matches on this map usually take very long and the players should have enough equipment then.

I think that `n = 70` might be a good compromise.

13_wall.conf
```diff
@@ -4,11 +4,17 @@ author = Thomas-S
...
 team.2.color = blue
 team.2.pos = (-51,-15,-51)
 team.1 = red
 screenshot = 13_wall.png
+chests.1.from = (-69,-16,4)
+chests.1.to = (69,23,69)
+chests.1.n = 70
+chests.2.from = (-69,-16,-69)
+chests.2.to = (69,23,-4)
+chests.2.n = 70
```
ClobberXD: New-line required here

ClobberXD:
Alright. Will approve once the line comment has been addressed.

Thomas--S:
done

PR #42, opened 2019-07-23
Repair plain battle

JostP:
Red side has only 4 mese (it should have 6). I have no idea how this happened :). So I added new 2 mese on red side.


JostP:
Improve tunnel commit :/ ? Sorry, will reopen this PR

PR #43, opened 2019-07-23
repair plain battle

JostP:
Red side has only 4 mese (it should have 6). I have no idea how this happened :). So I added new 2 mese on red side.

****

Successor of #42


ClobberXD:
The file name seems to be incorrect.

JostP:
> The file name seems to be incorrect.

incorrect?

ClobberXD:
It seems to be fine now o_O

I could've sworn it was `plain_battle.mts_10` before...

Thomas--S:
Labeling as WIP (see #50)

JostP:
Improved wall, different design. I made front row indestructible, other 5 are destructible so players can modify wall (behind indestructible row) as they want.

"Defence corridor" is still quite useless, because guns that we have aren't very useful for repeling enemies when they are moving. I hope "sniper mod" will make it more useful.

I also added shooter:pistol as spawn item to encourage players to use it.

PR is ready for review, and don't forget this PR was created to add 2 mese on red side.

And one screenshot:
Zaslonska slika 2019-09-20 17-23-11

MinetestSam:
Checked mese on both the sides and yes indeed there are 6 on both the sides but my only concern with my map right now is that although the front row is indestructible the top of the towers are still destructible which either way I am fine with but I think it should be made indestructible
screenshot_20190921_155336

JostP:
> Checked mese on both the sides and yes indeed there are 6 on both the sides but my only concern with the map right now is that although the front row is indestructible the top of the towers are still destructible which either way I am fine with but I think it should be made indestructible

indestructible wall is intended to prevent the invasion of enemys into "defence corridor", so I don't see why tower should be indestructible.

MinetestSam:
See https://github.com/MT-CTF/maps/issues/50#issuecomment-532352895

JostP:
Well, I will make towers indestructible, no problem, but I still think this isn't so neccessary.

JostP:
Towers are now indestructible

MinetestSam:
The `.conf `file seems to be missing the license.

JostP:
Added license

JostP:
License was actually already added for this map in #49, but I created this PR before adding a license.

PR #44, opened 2019-07-23
repair Karsthafen town

JostP:
Stonebrick slabs doesn't load on server, so lava flow. I repaired lighthouse

Zaslonska slika 2019-07-20 18-21-40


ClobberXD:
This is strange... I remember inspecting this lighthouse before approving the map, and it was intact. I tested the map in CTF, not MTG, btw.

Thomas--S:
I wasn't able to reproduce the missing slab error locally.

ClobberXD:
I can't reproduce this either. But I wonder how that happened in the server. It's an indestructible node, so no one could've mined it.

PR #45, opened 2019-08-07
Add Wool land

JostP:
World converted into wool (default:dirt --> wool:dark_green, default:stone --> wool:grey, etc.).
Zaslonska slika 2019-08-06 16-28-24
Zaslonska slika 2019-08-07 10-41-15
Zaslonska slika 2019-08-07 10-45-16


ClobberXD:
Interesting concept, actually - I quite like the idea. But the terrain is very "normal" and there's nothing else distinctive about this map, unfortunately. So, I'm not sure what is to be done here...

JostP:
> Interesting concept, actually - I quite like the idea. But the terrain is very "normal" and there's nothing else distinctive about this map, unfortunately. So, I'm not sure what is to be done here...
> 

Yes the terrain is very normal, but there is everything made with wool. This map is very colourful, players can make colourful buildings walls etc. that is why it is something special. They can also tunnel and mine with swords (ores weren't converted into wool). Maybe, should players get some wool as initial_stuff? 

> Btw, assuming that "improve tunnel" is the latest commit in your `master` branch, enter the following commands to remove it:
> 
> ```
>...
> ```
> 
> Or you could just delete the `mods/ctf/ctf_maps/maps`, and re-clone the repo. Make sure you've pushed all your branches to your remote, though.

Thank you! :) Will try to do this later, I am away this week.

ClobberXD:
> Yes the terrain is very normal, but there is everything made with wool. This map is very colourful, players can make colourful buildings walls etc. that is why it is something special. They can also tunnel and mine with swords (ores weren't converted into wool). Maybe, should players get some wool as initial_stuff?

Well, I agree the wool is the highlight of this map, but other than that the rest of the map is a plain boring map with no artificial structures or terraforming, to put it bluntly. I'm neutral on this.

JostP:
well, I will close this PR

What do you think about converting `river valley` or `forest map` into wool? We have a lot of "classic" maps such as 2H2L, POTP, coast of nostalgia, river valley and forest map.

ClobberXD:
If you really want a wool map, we could convert an existing map to consist of only wool. But I'm not sure which map it should be, among the "classic maps". Feel free to suggest the map that you think is the best candidate for conversion.

There needs to be a lot of discussion prior to the conversion, to make sure everyone knows and approves of it. @Thomas--S, your thoughts on this matter?

Thomas--S:
To be honest, I am strongly opposed to converting an existing map to wool. All of the classic maps feature a unique gameplay style which would be lost by converting them to wool.

Actually, I'm more in favor of adding a new map as a wool map, as proposed in this PR.

To prevent player being bored, I suggest to disable the maps that are most disliked by the players and to change the disabled maps every few weeks/months.

---

Regarding the proposed map in this PR:

The base of one team seems to be on a hill with no forest, while the other base is on flat land with a very near forest. The latter base disadvantages its team, I think, because enemies can approach much more easily.

ClobberXD:
Good points. I'm fine with a brand-new woolen map if redundant maps can be disabled at will.

JostP:
Added license

JostP:
Do you think that players should be given some wool at spawn?

ClobberXD:
> Do you think that players should be given some wool at spawn?

Good idea. It'd be nice to have x wool of each colour

JostP:
Is possible to get random?

ClobberXD:
> Is possible to get random?

No, that's not possible (at the moment).

JostP:
added 5 colours of wool (blue, red, orange, yellow, magenta) as initial stuff. 20 blocks each colour

MinetestSam:
The map doesn't have any iron ore and also there are couple of random cobblestone stairs near the red base at the bottom.

JostP:
There really isn't any iron ore? This is bug then, I will check and repair it.

MinetestSam:
Yes, I am pretty sure there is no iron ore but again you shouldn't go by my word and do a world edit test yourself.

Also please remove the random stairs.

JostP:
Is it good idea if I replace wool:black (gravel) and wool:white (silversand) with iron ore?
Replacing coal with iron would make too much iron ore. (or not?)

I really have no idea when did copper, iron and tin disappear. I have never replaced them with something else.

Zaslonska slika 2019-10-11 19-12-46

MinetestSam:
Here you can do two things,
1) You can add steel sword to initial items so that there would be no need to add iron ores
2) You can strategically place iron ores (like you did with mese in your Plain Battle map)

I think replacing black wool would also make quite a lot of iron ores and we don't know if that would be equal on both sides so it will just increase your work.

JostP:
Can I replace `ctf_map:stone` with `ctf_map:grey`(grey wool)?

JostP:
What about adding some mese in each mine?

JostP:
Done. Changes:
- 2 mines (under black fort), without meses
- forest tower near forest
- replaced `ctf_map:stone` with `ctf_map:grey` (barrier)
- added hint
- removed stairs

ClobberXD:
@JostP It's recommended to leave the barrier alone, for the sake of consistency. Feel free to use your creativity _within_ the barriers, but please refrain from modifying the barriers themselves. This could also potentially affect features that depend on `ctf_map:stone` being the outer barrier, leading to undefined behaviour.

JostP:
Replaced barrier back to stone, added new screenshot.

MinetestSam:
A few more changes are needed before I can approve. Since there is a mine now the players should spawn in with a stone pick at least and also few sticks since there are no trees. Also maybe change the hint to `Find the mine and mine ores.` (it is up to you, it is just a suggestion)

JostP:
Done: changes
- change hint
- add wood to mines and made ladders destructible (stick source)
- stone pick as initial stuff

JostP:
Please review hint. I don't know if it should be "Look for **the** mine on the map!"? Im not english native speaker :gb:

MinetestSam:
I would still prefer if the players spawned with sticks because the noobs can just mine the wood and players would be hanging in the middle of the match with no ores to mine.

Also I don't understand the idea behind the empty rooms and I don't think they provide any purpose.
screenshot_20191020_204207

JostP:
This room was created naturally on map, but I can remove it.

About sticks: If I add sticks as spawn item, I have to remove one color of wool. I really don't think there is not enough material for sticks, for example you can dig ladders and you get a lot of sticks, sticks spawn in team chest, you can still get wood in treasure chests (and I don't think noobs will waste it). At last there fewer than 50 steel so you need only 25 sticks.

MinetestSam:
You won't need to remove anything, there are currently 7 spots occupied in the inventory and the player inventory has 8 slots, so one slot is still empty

JostP:
I want to give players one empty slot so they can shift-click steel sword from chest and get it into hand, without pressing "I" and manually moving stone sword with steel.

JostP:
Do I have to remove this room if it is naturally generated? Many maps have rooms like this one.

JostP:
Update:
Removed "empty rooms", rename files to 20_wool_land.(png/mts/conf).

Players still don't get sticks (see previous comments for explanation).

Ready for review, everything is fixed now.

MinetestSam:
From what I have seen, I think players usually put swords in the first or second slot and I have rarely seen them using the 8th slot. Even if they are in a hurry and want to shift click the sword it would go in the 8th slot for which they would have to change the slot of the sword to their required or favourite slot.

JostP:
But still, why would players get sticks as initial stuff if there is no need for it? You can get sticks from team chest, treasure chests and from ladders and wood from mine. It can maybe happens once in 500 games that players (noobs) destroy all wood which exist on the map

I just don't want to give sticks as initial stuff because it will be pollution of the inventory.

If you don't agree, can we ask ANAND or Thomas-S about it?

ClobberXD:
IMO, sticks would be a good idea, *if* the hotbar wasn't already packed with a bunch of initial stuff. If there's enough sticks or resources for sticks in the map to last forever (I haven't checked yet), then sticks aren't needed.

MinetestSam:
I have suggested adding sticks because of 3 main reasons,
1. People (casual players and not pros or pros who are new to the game) focus less on crafting so they don't read what is exactly required for the recipe of an item which let's say is a sword.
2. I would rather have my chest filled with steel swords than having a slot just for sticks or wood which I would have to convert to sticks too.
3. Convenience: Players when respawned would only need to take out the iron and then could convert it into a sword and directly go in battle which is much faster.

JostP:
Ok, I give up, you convinced me. I will add sticks to inventory + give back one color of wool, which I removed once, so sticks will be in second row of the inventory.

JostP:
Done

ClobberXD:
I notice inconsistent use of indestructible wool in places like the dark grey watch towers, the watch tower in the middle, and the underground "ore rooms".

JostP:
> I notice inconsistent use of indestructible wool in places like the dark grey watch towers,

This towers are intended to be used by enemy, so exterior nodes are indestructible to prevent tower to be destroyed with guns (pistol destroy wool).

> the watch tower in the middle

Here isn't any indestructible node, is it? Here is no need for indestructible wool.

> and the underground "ore rooms".

Yes, these nodes shouldn't be indestructible, my inaccuracy. I'll fix that if you want to have them destructible.

ClobberXD:
> This towers are intended to be used by enemy, so exterior nodes are indestructible to prevent tower to be destroyed with guns (pistol destroy wool).

Ah, I see. Why not make them completely indestructible then?

> Here isn't any indestructible node, is it? Here is no need for indestructible wool.

That, is exactly my question. Why do you want to leave this tower breakable, given that you fortified the dark grey ones?
 
> Yes, these nodes shouldn't be indestructible, my inaccuracy. I'll fix that if you want to have them destructible.

That's up to you. I suggest either making them completely indestructible (I don't see the point of that, though), or completely destructible.

JostP:
Done:
- dark grey towers have no changes, players can modify their interior as they need.
- Watch tower is completely indestructible now,
- wool in mines are now destructible + I removed wood because players get sticks as initial stuff

JostP:
Thanks for merging and reviewing so many times

PR #46, opened 2019-08-07
Add Towers by Atmozpheric

atmoz:
Two towers and an air balloon with iron ore in the middle (also a crashed balloon on the ground with smaller iron ore).

Should be easy to defend against ground attacks, forcing the teams to attack from the air (building bridges). A right balance between ground and air defense is key to win on his map. Hopefully a variety of strategies are possible from match to match.

Map is temporary hosted on a CTF server: 3.120.138.233:30001 (or search for "towers" in server list).

Screenshot from 2019-08-01 19-47-55
Screenshot from 2019-08-07 19-10-04
Screenshot from 2019-08-07 19-11-32
Screenshot from 2019-08-07 19-16-09
Screenshot from 2019-08-07 19-17-38


ClobberXD:
I really like the idea of the team bases being on top of towers. :+1: 

Will review soon.

ClobberXD:
Tested. This map offers a very unique style of gameplay that no other map offers, thanks to the bases placed high up on the towers. I like the air balloon, and the crashed air balloon is a very nice touch.

OTOH, I feel that the air balloon in the sky is pretty high, and the small amount of iron ore isn't precious enough to warrant a space trip like that. Maybe make it more interesting, by maybe filling the insides of the balloon with more precious ores? Or the gondola could be made bigger to accommodate more iron ore.

Thomas--S:
> by maybe filling the insides of the balloon with more precious ores?

I strongly oppose the addition of mese and diamond on this map type because it will most likely lead to a very uneven gameplay with too many advantages for the defending team.

However, I support the idea of having large amounts of iron (e.g. as a treasure in the balloons) available on this map.

atmoz:
Currently there are 14 iron ore in the gondola and 8 iron ore hidden inside the balloon itself (entry on the top). On the ground is additional 6 iron ore. Not sure how much is enough or too much?

The fact that you can get steel swords from chests on the ground also lessen the motivation for going for the balloon. I wonder if it's possible to somehow spawn the chests in the balloon :thinking: But that would require much more space.

Thomas--S:
I'd suggest that even about 100 iron ores in total might be helpful for this kind of map. There won't be too much imbalance (because players can "only" obtain steel swords and nothing better), but steel can also be used for strategy, i.e. by using doors and spikes etc.

ClobberXD:
I agree, while there seem to be ample weapons in the chest, all the extra ore can be used for making other items like doors, traps, fortifications, etc.

JostP:
Are ladders indestructible? If they are breakable players will easily destroy them (with smg in 30 seconds for example). players won't be able to climb down and up and they will probably build up and make ugly 1x1 towers.
I haven't tested it so that is why I am asking.

ClobberXD:
All nodes in both the towers (including the ladders) are indestructible. So is the air balloon in outer space.

atmoz:
Now there are 27 iron ore in the gondola + 20 inside the balloon and 16 iron ore on the ground.

ClobberXD:
Nice. Will re-review in a couple of days.

ClobberXD:
Everything looks good, except for one teensy thing - doors. Doors don't work in CTF, so it'd be nice if they could be removed. Or maybe, the doorways could be sealed with normal nodes instead (wooden planks, for example)

atmoz:
Removed the doors

MinetestSam:
You also need to change the prefix number to your map to 19 now.

Thomas--S:
The prefix numbers can be changed on merge.

atmoz:
- [x] Screenshot in conf
- [x] License in conf
- [x] Prefix changed to 19

19_towers.conf
```diff
@@ -0,0 +1,17 @@
+name = Towers Beta
```
ClobberXD: Is this still in "Beta"?

ClobberXD:
The screenshot might benefit from a higher `view_range` IMO. It's too foggy to see much, at the moment.

19_towers.conf
atmoz: Oh, snap! :-)

atmoz: Fixed

atmoz:
- [x] Remove beta from name
- [x] Take new screenshot with larger view range

atmoz:
- [x] Replace normal blocks in outer map barriers with indestructible blocks

atmoz:
- [x] Replace `ctf_map:stonebrick` with `ctf_map:stone` under barrier
- [x] Add `ctf_map:stone` floor on -100

JostP:
Tested, everything works, everything what was requested is fixed :+1:

JostP:
One question: was naturally generated iron ore underground removed on purpose?

atmoz:
No, I haven't done anything, not even checked if there was any ores.

JostP:
> No, I haven't done anything, not even checked if there was any ores.

Asking because when I removed stone and dirt there was only copper, coal and tin ore. I didn't found any iron ore (underground)

atmoz:
Not sure if it's good or bad :thinking: It will certainly focus the teams on securing the iron ore in the balloons, I think that could be a good thing. But it will on the other hand also limit the possibilities if one team secures both balloons and takes the iron ore.

PR #48, opened 2019-08-27
Add "Big Ocean" map by gameit

JostP:
see #35 

Screenshot is missing


Thomas--S:
Please add license information.

JostP:
> Please add license information.

  and screenshot (400 x 600 image (for /maps command))

> ClobberXD changed the title Add "big ocean" map by gameit Add "Big Ocean" map by gameit 8 days ago

I don't have time to occupy myself with spelling on school holidays :)

ClobberXD:
> I don't have time to occupy myself with spelling on school holidays :)

No worries, I exist for a reason ;)

JostP:
UPDATE: https://forum.minetest.net/viewtopic.php?f=12&t=19729&sid=8aff5d9a81706e0dbdd7111eb3a2b8de&start=150#p358219

JostP:
We still need screenshot

ClobberXD:
:+1: for everything except for the inconsistency of using destructible nodes throughout the map but (unintentionally?) making the sails out of indestructible white wool. Will approve once that's fixed.

@JostP Here's a screenshot, which you can use, if you think it does justice to the map:

screenshot_20191002_152731

Thomas--S:
Please be aware that license information still seems to be missing.

JostP:
Well, there is also second (quite big) problem: Caves and ores are only on blue side.

Screenshot after replacing stone with air:
Zaslonska slika 2019-10-03 15-51-37

ClobberXD:
Should those be removed altogether?

JostP:
Quote gameit:
> Big Ocean 2.0
> After a time of thinking, I finished the new version of my map "Big Ocean".
>   - Changed the deep of the water from 10 to 20
>   - And placed ctf-stone in the middle of the map, so you can't build a tunnel from one base to the other
>   - Added a new bigger type of island
>   - Put some minerals/caves in the stone
>   - Added a skypart
>   - Placed more ships


Ores were added as a substitute of removing steel sword from spawn items. So ores maybe should stay if they are on both sides.

BTW Thank you for adding me to map reviewers team :D

JostP:
So what do you think, should ores stay or be removed? If they stay, should player get stone sword?

ClobberXD:
If gameit can add ores to the other side, then I'd personally prefer that approach. If not, I'd prefer to not have any ores, and give steel swords to players.

JostP:
Update:

Map now have:
- more than 300 000 iron
- more than 300 000 coal
- more than 300 000 copper :smile:

JostP:
What about if ores aren't removed? You get steel sword, so you can have extra steel to craft doors, traps etc. Wouldn't that be great, to have hundreds of doors in-game, bases built with rainforced stone or damage-cobble (both are not possible to craft many on normal maps)?

MinetestSam:
Yes it does provide for various items like traps and doors to be made but I highly doubt it would be used for that since crafting(crafting of traps and doors) is only used in Pro Vs Pro matches.

MinetestSam:
Players when they get steel swords usually don't tend to go for mining of resources like iron because their utility of steel sword is already fulfilled so they don't need to go for mining. Ice Age is a good example here, the map has I think 99+ iron ores yet players don't go mining and they like to engage directly in combat as there is no incentive to go and mine.

JostP:
Just for information, what means 300 000+ blocks of each ore :)
Zaslonska slika 2019-10-24 10-09-42

JostP:
added license + screenshot line

big_ocean.conf
```diff
@@ -1,12 +1,14 @@
...
-team.1.color = red
+rotation = x
 team.1 = red
+team.1.color = red
+team.1.pos = (-1,-9,100)
+team.2 = blue
+team.2.color = blue
+team.2.pos = (-1,-9,-100)
+initial_stuff = default:pick_steel,default:sword_steel,shooter:pistol
+license = GNU General Public License
```
ClobberXD: The GPL license's version is missing

JostP: Quoting gameit's LICENSE file:
 ```
    ...
    This program is free software: you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
    the Free Software Foundation, either version 3 of the License, or
    (at your option) any later version.
    ...
```

Does that mean version 3?

ClobberXD: Correct

ClobberXD: This, however, makes me wonder if it's wrong to just display the name. I think we should allow map makers to add the license name, and a file containing the full fat license deed.

ClobberXD:
> Yes it does provide for various items like traps and doors to be made but I highly doubt it would be used for that since crafting(crafting of traps and doors) is only used in Pro Vs Pro matches.

While that's somewhat true, I don't see a problem per-se, in just letting the ore be.

JostP:
Added license version

big_ocean.conf
MinetestSam: But gameit did have a version file in the downloads which stated it was version 2.2 now I am wondering whether that is the version of the "map" or the licence (It shouldn't be licence since there is no v2.2 I think)

Thomas--S:
To be honest, I'm not sure if we really should allow GPL licenses for maps. They are intended for source code, but maps are media files, so I guess that there could be problems when applying GPL to maps. (What is the "source code" of a map?)

Please note: I am not a lawyer.

I strongly suggest not to merge this map until all open questions regarding licensing are solved.

big_ocean.conf
JostP: I think it is map version, but we can still ask him in-game when he appear

ClobberXD: There isn't a v2.2 version of the GNU GPL. It should refer to the map itself.

ClobberXD:
> To be honest, I'm not sure if we really should allow GPL licenses for maps. They are intended for source code, but maps are media files, so I guess that there could be problems when applying GPL to maps. (What is the "source code" of a map?)

Disclaimer: IANAL

"Source code" refers to the actual data that makes up the licensed work. In this context, the schematic file is the source code.

https://www.gnu.org/licenses/gpl-faq.html#GPLOtherThanSoftware
 
> I strongly suggest not to merge this map until all open questions regarding licensing are solved.

Apart from whether using GNU GPL v3.0 is a good choice for non-software works, the only legal-ish issue that prevents this PR from being merged:

> This, however, makes me wonder if it's wrong to just display the name. I think we should allow map makers to add the license name, and a file containing the full fat license deed.

https://github.com/MT-CTF/maps/pull/48#discussion_r339414730

JostP:
Added torch as initial stuff

MinetestSam:
Yeah I think the source code here is the schematic file since it is the "main" thing the map creator created. 
Quoting from the GPL site,
> You can apply the GPL to any kind of work, as long as it is clear what constitutes the “source code” for the work. The GPL defines this as the preferred form of the work for making changes in it.

JostP:
Maybe we should still ask gameit about license version before merging

ClobberXD:
@JostP Did you add 300000 of each ore?

Thomas--S:
> The GPL defines this [the source code] as the preferred form of the work for making changes in it.

I am not sure if the schematics are really "source code" with this definition. At least I prefer to make the changes to the actual Minetest world (which is not available) rather than directly modifying schematics.

ClobberXD:
> I am not sure if the schematics are really "source code" with this definition. At least I prefer to make the changes to the actual Minetest world

But you publish the changes via the schematic itself, right? It's like saying that you don't modify the code directly, but you only modify the copy of the file that your text editor has loaded. :)

(Disclaimer: IANAL)

Thomas--S:
As always, I am not a lawyer.

But to me it seems like distributing only the schematic is like distributing the compiled program. It is theoretically possible to make changes to the machine code (i.e. schematic) directly, but most developers prefer to make the changes to the source code (here the Minetest world).

ClobberXD:
> it seems like distributing only the schematic is like distributing the compiled program

Unlike a compiled program in machine code, a schematic is just a serialised version of a Lua table, which can be de-serialised either manually, or by using `minetest.read_schematic` (5.1.0+ only, IIRC).

MinetestSam:
Good examples given here about GPL for non-software https://opensource.stackexchange.com/questions/741/with-gpl-what-does-source-code-mean-for-non-software

Thomas--S:
From MinetestSam's link:

> Think of 'source' as something which is created in a manual process and then fed into an automated process to produce the final artifacts.

As far as I can understand, the Minetest world is shaped in a manual process and then fed into an automated process (i.e. map exporter) to produce the final artifacts. (i.e. schematics).


---

In future, please don't merge maps as long as staff members have worries regarding the licensing. Having to wait a few days longer until open questions are discussed is a lot less of a problem than having legal problems regarding licensing.

I'd strongly suggest to revert this merge commit.

Please note: I am not a lawyer.

ClobberXD:
> As far as I can understand, the Minetest world is shaped in a manual process and then fed into an automated process (i.e. map exporter) to produce the final artifacts. (i.e. schematics).

Am I correct in assuming that you seem to be agreeing to the use of GPL for this map?

Taken from https://opensource.stackexchange.com/a/742:

> The source code for a work means the preferred form of the work for making modifications to it.
What's the preferred form for making modifications? You might deduce what your project's source code is by going through the following sequence:
>
>    - Consider the work that is supplied to the recipient.
>
>    - Consider a range of possible changes that recipient might want to make: move images around, give a character different movements in an animated movie, alter text, change the order of sections in a document, bring obscured background elements to the foreground unobstructed, change a single instrumental track in an audio file, etc.
>
>    - What format of the work would make those editing tasks easiest?
>
> Usually, the answer will be obvious: it's the layered Photoshop or GIMP file you used to create your finished image, or the word processing document you use to create your PDF output, or the multi-track audio-workshop file you compressed down into only a two-band stereo output. For a web page with images, the source code would be non-obfuscated HTML and the most-easily editable versions of the images.
>
> Sometimes (I'd guess rarely) the answer is not obvious. For example, the Battle for Wesnoth community has had a non-trivial discussion about whether MIDI or OGG constitutes the preferred form for making modifications, because MIDI is closer to editable "source" but exactly how a MIDI will render to a proper sound file will vary by MIDI environment. Aside from strange edge cases like this, it should be obvious what document format is most appropriate for making modifications.
>
> Other noteworthy cases:
>
>    - If the work has always been in a form that is not friendly to making modifications (e.g., a complex image made in a simple Paint-style program without layer information, or a binary executable written by hand by a very clever hacker with a hex editor) then that is the best "source" that exists and is sufficient: that was, apparently, the original author's preferred format for creating and modifying the work.
>    - Conversely, if the project has always been in a format most conducive to making modifications (e.g., human-readable HTML for a website), then the work is composed entirely of source code.
>    - For the simple-image-in-Photoshop hypothetical, if the Photoshop file has metadata what would make it easier for a person to perform edits, that should be preserved: e.g., if you specified gradients/regions for color alterations and that information is available as separately-editable layer-like feature, then certainly that would be preferred for making modifications. If the transformation you did merely altered pixels and has no associated metadata or layer information, then the Photoshop file affords no advantage over the raw image in terms of ease of modification. However, unless I were absolutely sure that no such editing-friendly metadata existed, I would err of the side of caution and distribute the PSD file as well.
 
GPL certainly isn't the best license for non-software works, but it should work just fine as far as I can see. Of course, we can also ask the author to reconsider their choice of license for the map.

Disclaimer: IANAL

Issue #47, opened 2019-08-09
License of old maps


Thomas--S:
Many of the map files don't have a license added in their `.conf` file.

I strongly suggest to add an appropriate license to all maps.

I think we should only merge new maps if a valid license is stated in the `.conf` file.

CC BY-SA 4.0 might be a good idea for most maps, I think.

Please note: I am not a lawyer.

I ask all map creators to either submit a PR with the change in their `.conf` files or to state in this discussion what license they want to use so that the map team can edit the files.

We'd need an answer from the following map creators:

@rubenwardy 
@ClobberXD 
[WHYjustbecause](https://forum.minetest.net/memberlist.php?mode=viewprofile&u=19630)
@xenonca 
@JostP

ClobberXD:
I'm fine with CC BY-SA 4.0 for IceAge

rubenwardy:
CC BY-SA 4.0 is good by me

JostP:
CC BY-SA 4.0 for my maps

ClobberXD:
We should enforce a fallback license (CC BY-SA 4.0, for example) for maps that don't explicitly specify a license.

Thomas--S:
I am not a lawyer, but I doubt if this was sufficient.

For example, it is theoretically possible to create a map without using the map maker GUI, so the information about the fallback license would be missed.

Instead, I suggest to add a input field for licenses to the map maker GUI, so most maps that are submitted will automatically have an appropriate license.

ClobberXD:
That's exactly what I suggested. A license field should be added, but there should also be an explicit mention of the fact that if no license is specified, CC BY-SA 4.0 is assumed.

Thomas--S:
I simply suggest to only enable the export feature of the map maker GUI as soon as a license is specified. This way we "force" the map maker to actively choose a license.

I think that the license choice should be an active decision by the map maker and no fallback should be assumed.

ClobberXD:
Oh nice, I like your idea better.

xenonca:
CC BY-SA 4.0 is ok for Swamplands and Ancient Pyramids too

ClobberXD:
@xenonca "Ancient Pyramids" hasn't been merged yet, so please add `license = CC BY-SA 4.0` to the map's conf file.

Thomas--S:
#49

PR #49, opened 2019-09-06
Update license information according to #47

Thomas--S:
Remove Desert Spikes because of unclear license


README.md
```diff
@@ -3,4 +3,6 @@
 Feel free to create PRs for new maps.
 Also see the [Map Review](https://github.com/MT-CTF/capturetheflag/wiki/Map-Reviews) wiki page
 
-License: Varied
+## License
+The individual maps are published under various licenses.
```
ClobberXD: I suggest re-writing this section a bit to something like:

```markdown
## License

All individual maps are published under their own license, as given by the `license` field in `<map_name>.conf`
```

PR #52, opened 2019-09-27
Add Abandoned Isles by gimp

MinetestSam:
Continuation of #26


18_abandoned_isles.conf
```diff
@@ -0,0 +1,14 @@
...
+team.1 = red
+author = gimp, caoutiouspotato
+rotation = x
+h = 140
+team.2 = blue
+name = Abandoned Isles
+initial_stuff = default:sword_steel,default:pick_steel,shooter:pistol,default:sandstone 50
+r = 115
+screenshot = 18_abandoned_isles.png
+license = CC-BY-4.0
```
ClobberXD: Hyphens should be removed: https://creativecommons.org/licenses/by/4.0/deed.ast

```suggestion
license = CC BY 4.0
```

MinetestSam:
I kinda don't want a lot of chest to spawn on the map so that the chances of getting a shotgun are quite low because a shotgun or SMG on this map would be quite destructive.

ClobberXD:
> I kinda don't want a lot of chest to spawn on the map so that the chances of getting a shotgun are quite low because a shotgun or SMG on this map would be quite destructive.

- All islands and their connecting paths are indestructible.
- Treasures can be defined per-map, given which the default set of treasures won't be registered for the map.

MinetestSam:
Tested and from the looks of it, everything works. My only concern is that the water can allow players to float above but that really doesn't give any advantage to them. 
screenshot_20191004_122635

PR #59, opened 2019-10-05
remove doors and furnaces in "karsthafen bridge" map

JostP:
Tested, works


JostP:
Removed also `doors:hidden` in second commit. What actually is door:hidden?

ClobberXD:
`door:hidden` is an unpointable invisible node used to occupy the pos right above the door's base pos. This is to prevent falling nodes to fall into that pos, and get stuck until the doors are removed (IIRC).

PR #62, opened 2019-10-20
fix heretic mistake

afshaan4:
**sword, gun, pick, sand**

# why this is needed

* so you can swap between gun and sword

* so you don't need to suffer

* it grants you levitation

* I made the map

## This is a mistake
heretic


MinetestSam:
There is actually no need to that but I guess,

a. You are the map creater
b. I think most people will prefer having pistol in the second slot

PR #63, opened 2019-10-28
Add "maze" map by Waterbug

JostP:
https://forum.minetest.net/viewtopic.php?f=12&t=19729&p=360497#p360445

This map need a license (like all Waterbug's maps) and a screenshot. Screenshot can be made by me if Waterbug doesn't post any.


JostP:
Tested:

- Map is completely indestructible.
- There are ctf_map:wool_wool_wool_violet (whatever this is :smile: ) nodes on the map.
- I'm not sure about eligibility of this map, as you can see on screenshot below it is quite complex

Here are some screenshots of the map:

Zaslonska slika 2019-10-28 19-22-18

Zaslonska slika 2019-10-28 19-18-02

Zaslonska slika 2019-10-28 19-18-30

JostP:
Map looks promising :+1:

JostP:
Added screenshot

ClobberXD:
Map looks interesting at first glance. :+1: 

****

@JostP Thanks for putting in time and effort to convert ZIP files into PRs; much appreciated.

JostP:
Added license `license = CC BY-SA 3.0`

JostP:
`ctf_map:wool_wool_wool_violet` need to replaced with `ctf_map:violet` and then PR is ready to go

JostP:
Also make barriers with `ctf_map:stone` (exterior nodes)

JostP:
Update, but map still need one small change: https://forum.minetest.net/viewtopic.php?f=12&t=19729&start=175#p360823 .

JostP:
Tested LGTM :+1: 
One more approval needed by ANAND or Thomas--S.

ClobberXD:
Will test today, if possible

ClobberXD:
It's very unconventional but engaging - I really like the map. However, I notice that some jumps are outright impossible to perform without engaging sprint, and this will deter Android players from playing on the map.

Editing the map isn't the solution IMO - editing is tough, and the map would look ugly if it isn't completely re-built from ground up. OTOH, I'll finish implementing per-map physics modifiers and create a PR tomorrow. Once that PR is merged, we can tweak the `jump` value in this map, to allow players to jump higher and further.

JostP:
Fixed directory structure, removed screenshot line.

MinetestSam:
Physics can now be changed so removed the label

JostP:
How much do you sugest jump should be? Feel free to push your commits to this PR, I'm away today and tomorrow.

ClobberXD:
It requires careful trial-and-error, but I suggest `phys_jump = 1.3` as a starting value.

ClobberXD:
@JostP Does your approval still hold?

JostP:
I approved map, but I will test it again with new jump value.

JostP:
Added line phys_jump = 1.3

I haven't tested on android, actually I have never played minetest on android, so someone with more experience should test new 1.3 jump and approve this.

Maybe we should also add `phys_gravity = something`?

ClobberXD:
Tested with a jump modifier of 1.3 - it's possible to jump over all ledges without sprint now, but jumping in corridors makes the player hit the (3-node-high) ceiling, making it an annoying experience. Then I tried a speed modifier of 1.2 instead - this time I was able to jump across the chasm from the ladder to the corridor just as well. It was actually more consistent, because the player doesn't hit the ceiling and so doesn't get rebounded abruptly.

Another advantage of the slightly-increased speed specific to this map - if a player gets stuck somewhere far from base, they can't just dig back to base, and have to retrace through the labyrinth, which can be a very time-taking process if not for the speed boost.

Will approve once `map.conf` has been updated.

maze/map.conf
```diff
@@ -0,0 +1,14 @@
...
+team.1 = red
+author = Waterbug
+rotation = x
+h = 76
+team.2 = blue
+name = The Maze
+initial_stuff = default:sword_stone,default:pick_stone,default:torch 10
+r = 61
+license = CC BY-SA 3.0
+phys_jump = 1.3
```
ClobberXD:
```suggestion
phys_speed = 1.2
```

ClobberXD:
> I haven't tested on android, actually I have never played minetest on android, so someone with more experience should test and approve this.

I mentioned Android because while it's possible to sprint, it's a little hard to sprint using the touchscreen controls while climbing and moving forward at the same time. To test, just try to make all the different types of jumps without using sprint.

ClobberXD:
@JostP Since there have been changes to the PR after your approval, it's advisable to re-test the PR (at least the changed parts) and ensure that everything still works okay.

JostP:
I think we should have chest zones only in 2 big rooms (where trees are) because point of map is to find these rooms and get better weapons IMO.
Two small (like 10 chests) chest zones should ofc be at base rooms.

ClobberXD:
Sounds good to me

ClobberXD:
Dismissing @JostP's approval, as a lot has changed since then (and more changes are to come).

JostP:
Done
- added 7 chest zones, see comments where
- added hint

Tested, but chests didn't appear, no idea why. I tried multiple times. Can someone please review `.conf `file and tell if there is a problem?

maze/map.conf
```diff
@@ -1,14 +1,51 @@
-team.2.color = blue
-team.2.pos = (-24,2,-48)
-team.1.pos = (24,2,48)
-team.1.color = red
-team.1 = red
+name = The Maze
 author = Waterbug
+hint =
```
ClobberXD: Please don't use invisible ink for map meta. CTF can't read stuff written in invisible ink, and neither can I.

JostP: I forgot to add hint in first commit, but it was corrected in second commit.

JostP:
Fixed, problem happens when CTF rotates schem and coordinates aren't correct anymore (I was making chest zones in MTG, not CTF).

Tested, everything works.

PR #69, opened 2019-11-02
Add Isles of Skill map by DELTA_FORCE

TSafa-23:
Adds a map where it has been specially designed to be as fair as possible. It is centered around combat, and it has been made to discourage skybridging (while allowing players to bridge under the isles for interesting gameplay). For anti-griefing measures, nearly every single block you see (besides the middle isle) is indestructible. 
Screenshot from 2019-11-02 09-31-44
Screenshot from 2019-11-02 09-35-30
Screenshot from 2019-11-02 09-35-43
Screenshot from 2019-11-02 09-36-05

(The weird white speckled node you see along the wall is killnode to prevent players from walking along the wall as easily)


KaylebJay:
I'll need to try it out in singleplayer, but I don't much see the point of this map. Looks like just another prototype of Abandoned Isles or Rubenwardy's map (Serpent's Pass), and both have more validity than this map (from what I am seeing). 
Is the point of this map to kill everyone that does not know how to jump two blocks? :D

TSafa-23:
No, the goal is having fair playing grounds where skill determines the winner.

MinetestSam:
The map doesn't promote any strategy and isn't good for mobile players. With the knockback being a feature for now griefing and players dying in lava would be regular. It is a Won't Add from me but I'll like see what the other map reviewers have to say about this map.

Thomas--S:
From what I can see form the screenshots, I also support the "Won't add" position.

KaylebJay:
If you wanted *fair* playing grounds, you might like the map Plain Battle. But if this is your definition of fair, the only map that is less fair is Abandoned Isles.
I think lava will be the winner every time without any skill at all. Thanks to knockback, should be pretty easy to knock a player into the seething red.
Also, aren't those gold blocks? I don't think gold blocks are supported in CTF.

KaylebJay:
MinetestSam, I agree that knockback could be used for griefing. I hate that you can knockback your teammates, this really needs to be fixed. But completely offtopic anyways :-)

TSafa-23:
Knockback barely knocks you back at all atm, unless there was a change I didn't see.

afshaan4:
> Adds a map where it has been specially designed to be as fair as possible

lemme fix that for you:

"Adds a map where it has been specially designed to be as boring as possible"

* no terrain
* no places to shoot from
* no places to ambush from
* only one obvious path to the other flag

people need to stop making low quality flat maps

MinetestSam:
It's a "Won't add" so closing the PR.

PR #70, opened 2019-11-05
Improve "river valley"

JostP:
Small improvement.
Added 4 watch towers on hills, can be used by snipers who will use snipers from incoming sniper mod.
Tested, works
Screenshot:
Zaslonska slika 2019-11-05 09-18-36


JostP:
There's no need to hurry with reviewing, but it would be nice if we have this map in CTF v2

JostP:
Thanks, this PR needs only one approval, doesn't it?

MinetestSam:
I think it requires two approvals as it isn't a minor enhancement (I may be wrong here).

Thomas--S:
A PR like this usually requires two approvals. I try to review this next weekend.

ClobberXD:
I agree with MTSam - I'm not too happy about the same tower being used in multiple maps. I'm OK with the concept of a watchtower, but I'd really like to see a tower with a different blueprint or different nodes used.

JostP:
Ok, I will change them

MinetestSam:
Directory change needed.

JostP:
Update: new watch towers design, fixed directory structure 
Tested, ready for re-review

JostP:
continuation: #75

Issue #71, opened 2019-11-07
Fix missing barriers underwater on older maps


ClobberXD:
Should be pretty simple.

Maps confirmed to be affected by this bug:

- Two hills and two lakes
- Bridge

Maps that are _probably_ affected by this bug:

- Plains of the past
- Coast of nostalgia

JostP:
see PR #76

ClobberXD:
The issue is valid until the PR is merged.

JostP:
Ah ok, sorry

PR #76, opened 2019-12-04
Fix missing barriers underwater and holes in red barrier

JostP:
Closes: #71 

Maps affected by this bug (all rubenwardy's): 
* Two hills and two lakes
* Bridge
* Plains of the past
* Coast of nostalgia

I fixed red barrier underwater and replaced `ctf_map:ind_stone` with `ctf_map:stone` because MT world does not load `ctf_map:ind_stone` when you do //mtschemplace.


JostP:
Done
Fixed:
- red barrier underwater and 
- replaced `ctf_map:ind_stone` with `ctf_map:stone` because MT world does not load `ctf_map:ind_stone` when you do `//mtschemplace`.

Tested, everything looks good. Ready for review

Thomas--S:
> replaced ctf_map:ind_stone with ctf_map:stone

I'm not sure if this is a good idea. Afaik, we decided that ind_stone is the better option in some/many cases.

ClobberXD:
@Thomas--S `ctf_map:ind_stone` is now just an alias of `ctf_map:stone`. I think you have something else in mind.

Thomas--S:
I thought about the red barrier stone, but that's probably a different node.

MinetestSam:
Found two holes in Plains Of The Past barrier
screenshot_20191205_130928
screenshot_20191205_130954
Also I'm pretty sure there holes on Ice Age too due to the snow and also below the middle due to water

JostP:
Thomas--S I didn't replace any red barrier stone.

MinetestSam this holes happened because when map was created there was grass. So when CTF starts it replace grass with air and hole appear.
But if neccessary I can of course fix this holes.

JostP:
I will check for IceAge if it needs fixing

ClobberXD:
@JostP Feel free to fix the cheese holes in another PR. If you even want to plug those holes, ofc...

JostP:
Holes looks quite ugly on some maps, so will add holes fix in this PR. I went through all maps:
- IceAge needs fixing red barrier underwater and holes
- Forest map needs fixing holes
- POTP needs fixing holes
- Towers needs fixing holes

WIP, I will finish PR probably after weekend.

JostP:
Done.
Ready for review

MinetestSam:
Semi-approval, only one map (Karsthafen Town) is left which needs its barriers fixed.
screenshot_20191207_160216
screenshot_20191207_160244

JostP:
Will fix this one too, thanks for your thorough review MinetestSam :)

JostP:
Done, fixed holes in barrier on karsthafen town map.

PR #78, opened 2019-12-04
Add an ammo pack as initial item on abandoned isles

MinetestSam:


ClobberXD:
What's better than adding an ammo-pack as part of `initial_stuff`? Adding two of them, of course! :D

MinetestSam:
There were mixed replies, Xenon wanted only one ammo pack on the map. So open to debate and discussions.

JostP:
I think 1 is enough IMO as pistol usually doesn't even need an ammo (you die faster than spending all gun's life (life or however it is called) ;) )

Thomas--S:
I didn't play on this map yet, but I'd suggest to start with adding one and see if it is enough. A second one can be added later if necessary.

JostP:
Tested, works. Should I merge or wait discussion of adding one ammo to end?

ClobberXD:
He he he :smiling_imp:

PR #79, opened 2019-12-17
Set chest zones on Towers

atmoz:
chests-bottom
chests-middle
chests-forest


towers/map.conf
```diff
@@ -14,3 +14,33 @@ team.2.color = blue
 team.2.pos = (0,35,-63)
 
 initial_stuff = default:pick_stone,default:sword_stone,default:cobble 20,default:torch 5,default:stick 8
+
+# bottom of towers
```
atmoz: I just guessed the synax for comments, can someone verify?

rubenwardy: this is correct, it's the same format as minetest.conf

MinetestSam:
Neat use of custom chest zones 👍

PR #80, opened 2019-12-24
Add "Burning sands map" by Waterbug

JostP:
https://forum.minetest.net/viewtopic.php?f=12&t=19729&start=175#p363745


JostP:
Marking as `action/change needed` because screenshot is missing.

JostP:
Screenshot added. Ready for review.

MinetestSam:
Map is very empty and I have mixed opinions regarding the map.

samuelstegerKOLLEGAH:
The barriers and everything else work very fine but, a bit boring tho... and looks a bit like the boring old sand map which was removed ...

JostP:
Some buildings (a small towers somewhere or small cowboy village...)  and cactuses should be added, then I will approve this map.

samuelstegerKOLLEGAH:
Cowboy Villages are a very good Idea

MinetestSam:
> Some buildings (a small towers somewhere or small cowboy village...) and cactuses should be added, then I will approve this map.

I don't like buildings in maps and feel like this map is rather better to be left plain.

JostP:
So MinetestSam do you think this map is good enough to be added? I remember there was a discussion on discord about this and most people wanted it to be added.

Btw, you have probably heard that ANAND had left minetest permanently?

JostP (commented):
Looks good

PR #85, opened 2020-01-19
rearrange iceage's init items

afshaan4:
Greetings council, tis I Xxweebmaster420xX.

I have rearranged the init items for council member ANANDS map "ice age" such that the gun and the sword are directly next to each other, this I believe shall make switching weapons easier.


ClobberXD:
I appreciate you wanting to make quality of life improvements, but everyone has a different way of arranging stuff in their inv. and hotbar, and it's not worth fine-tuning trivial stuff like this. Sorry, but :-1:

afshaan4:
big sad

PR #88, opened 2020-04-15
Add Ancient Pyramids [v2]

xenonca:
Basically a revival of #32.
Added chest zones (@ClobberXD >.<) to the 2 empty pyramids so there is now something to explore.
The only issue I'm having is that I'm not sure which weapons to give as initial items. Firstly because of the new crafting system where recipes seemed to be going to change and secondly because of classes which will have their own specific items for each class. Maybe treasures could be a solution?


ClobberXD:
:D

xenonca:
Idk if this PR needs a full review since it's basically the old map with the requested changes, which can be review in the old issue (#32).

xenonca:
Boing

ClobberXD:
> The only issue I'm having is that I'm not sure which weapons to give as initial items. Firstly because of the new crafting system where recipes seemed to be going to change and secondly because of classes which will have their own specific items for each class. Maybe treasures could be a solution?

You're correct in saying that it's now up to classes to decide what weapon to provide to players. Currently, maps are best left to only determine what other *tools* players need (like shovels for the moon map), and "bonus" weapons (like dem bling swords, or coolio guns). The former is best provided as initial items, while the latter should be provided as treasures.

I think reviewing this map is best left to @JostP or @MinetestSam for now, as the others are busy with the upcoming v2.0 release. I would be glad to review and provide the second approval after the release though.

samuelstegerKOLLEGAH:
Map is playable, unique and kind of enjoyable and has no broken barriers, but I think the map would be more enjoyable if u make some more little buildings at the sides of the map because until now i think the map is too open too easy for snipers and at the sides it looks kind of boring. And one Question how many Diamonds did u add for each team in total ?

MinetestSam:
Hello, I am busy with real life stuff and won't be able to review stuff anymore. Sorry for the inconvenience caused by me.

JostP:
Reviewed. 
Map works perfectly. Points from ANAND's review in PR v1 are fixed. Chest zones are nice addition :+1: . The only thing that bother me is that map is very plain, but since this was fine by ANAND, this is not a problem. 

There is only one thing left to be fixed i.e. [screenshot size](https://github.com/MT-CTF/capturetheflag/tree/master/mods/ctf/ctf_map/map_maker#screenshot). Screenshot is used for /maps in-game so it should have ratio of 3:2. When this is fixed, maps can be added.

xenonca:
Replaced the screenshot with a 3:2 (600x400px) image

xenonca:
Bump

MinetestSam:
Screenshots shouldn't necessarily be 3:2 but it is preferred that they are.

PR #91, opened 2020-04-30
Remove shooter class items from initial items

MinetestSam:


abandoned_isles/map.conf
```diff
@@ -8,6 +8,5 @@ rotation = x
 h = 140
 team.2 = blue
 name = Abandoned Isles
-initial_stuff = default:sword_steel,shooter:pistol,default:pick_steel,default:sandstone 50,default:torch 10,shooter:ammo
-r = 115
```
ClobberXD, line 12: Please revert this change

abandoned_isles/map.conf
```diff
@@ -8,6 +8,5 @@ rotation = x
 h = 140
 team.2 = blue
 name = Abandoned Isles
-initial_stuff = default:sword_steel,shooter:pistol,default:pick_steel,default:sandstone 50,default:torch 10,shooter:ammo
```
ClobberXD, line 11: Why has `shooter:ammo` been removed?

MinetestSam: I thought it was a part of shooter class, my bad

ClobberXD: It would be a part of the upcoming sniper class, but there's no guarantee that the PR would be merged anytime soon. It's advisable to not pre-emptively code in anticipation of a future change.

Issue #93, opened 2020-05-13
Karsthafen town and bridge are identical


ClobberXD:
I've felt this numerous times myself, and most other players feel the same thing. Karsthafen town and Karsthafen bridge are way too similar to each other. The map regions literally overlap each other in the original Karsthafen world, and can be easily noticed while playing - I myself have gotten confused because of this, take my usual path, and end up hundreds of nodes from both bases before realising that this isn't the other Karsthafen map.

CTF maps should offer varied paces and styles of gameplay, and the two Karsthafen maps in question violate this principle. I'm all for removing either one of the violating maps. Feel free to discuss which one could be removed, and/or maybe provide alternatives to removing the maps.

I'd ideally prefer to resolve this issue before the great grand release of CTF v2.0, but after release is fine by me too, if that's what it comes down to.

JostP:
I agree that they are too similar to each other. I'm fine with removing one of them.

ClobberXD:
Good choice. Now onto discussing *which* one of the two should be removed.

What are the pros and cons of both maps? As the Karsthafen maps are based on a "city" theme, I'd suggesting removing the one that has less "city"-like environment and has more natural terrain.

JostP:
I think bridge one has less "city". I will test both maps once today.

Issue #84, opened 2020-01-03
Revert door-removing commits


ClobberXD:
Wooden doors in maps are supported after https://github.com/MT-CTF/capturetheflag/commit/05fba3f5cee7f1064b3ffe7a8760f609f8e769ac.

Note: Reverting the door-removing commits isn't as straight-forward as it might seem, especially if the schem was modified again. Or so I think...

Thomas--S:
I guess it might be the best solution to manually place the doors again while manually comparing the map to an older version.

PR #94, opened 2020-05-25
Remove karsthafen bridge map

JostP:
Closes issue #93 and #84 

EDIT:
Changes in this PR:

- remove karsthafen bridge
- rename karsthafen town to karsthafen
- re-add doors ( #84 )
- improve blue base's position
- new screenshot
- 50 chests on each side instead of 70


ClobberXD:
I think bridge has buildings in a more distributed fashion, while town has fewer buildings clumped together in the middle. The screenshots seem to confirm that as well. Also, the blue base in town map is at a big strategis disadvantage due to the various access points around, compared to the more guarded red base. Given these points, I'm more inclined to see the town map being removed.

@JostP, if you're not satisfied with bridge, we could replace it with another map of Karsthafen, this time with a lot of buildings.

samuelstegerKOLLEGAH:
Yes i think its better to remove it.

JostP:
Rebased.
I tested both maps again and I insist on removing karsthafen bridge. I noticed problem on it that there is something like only one way (one main street which leads from first to second base). Contrary karsthafen town has more branched streets and consequently more ways that leads to enemy base. Also it is more sniper friendly.

> Also, the blue base in town map is at a big strategis disadvantage due to the various access points around, compared to the more guarded red base.

Yes this is true and I am going to move blue base a little.

JostP:
Finished:
Changes in this PR:

- remove karsthafen bridge
- rename `karsthafen town` to `karsthafen`
- re-add doors ( #84 )
- improve blue base's position
- new screenshot
- 50 chests on each side instead of 70

JostP:
Tested works, merging...

PR #98, opened 2020-06-13
Wool map:remove flowers and some wool from inv

### LINKED ISSUE #97: wool_land: Provide many of a single node, instead of a little of many nodes
This frustrates me a lot, as I have to waste more time sorting my inventory than usual, because some of the most important items end up completely outside my hotbar - this is because my hotbar is already filled with various colours of wool, which I personally consider as unimportant, compared to more crucial items like weapons and meds. I also have to switch slots when one colour of wool has been used up, which is also time-taking, compared to having more of one single node.

I suggest giving only one colour of wool in a large quantity,
- to save inventory space
- to save inventory restocking time
- to also save yo' *insert word for "bottom-side" here* mid-battle by not needing to open your inventory while nerd-poling. :P

JostP:
(This pull request was generated by a script)

Closes #97


wool_land/map.conf
```diff
@@ -11,4 +11,4 @@ r = 115
 h = 140
 rotation = z
 license = CC BY-SA 4.0
-initial_stuff = default:sword_stone,default:pick_stone,default:torch 5,wool:blue 20,wool:red 20,wool:orange 20,wool:yellow 20,wool:magenta 20,default:stick 5
+initial_stuff = default:sword_stone,default:torch 5,wool:blue 20,wool:red 20
```
ClobberXD, line 14: *More* of the same node would be nice. I mean, increase the item count as well. :P

JostP: Would 2x30 also be good? Instead of 5 you now get only 2 colors of wool. Stone pick and sticks were also removed since there is no need for mining anymore

ClobberXD: > Would 2x30 also be good?

I'd personally prefer only one colour, but I guess I could adjust up to 2. The number of each colour of wool doesn't depend on the total amount of wool provided. The main reason for wanting more per-stack is:

> I also have to switch slots when one colour of wool has been used up, which is also time-taking, compared to having more of one single node

So even if you provide 2 colours of wool, I'd like at least 40 per-colour.

> since there is no need for mining anymore

I thought there were ores? If there's a possibility of mining, I suggest providing the stone pickaxe. It's the player's choice to mine or not to.

JostP: They can get a pick from chests, so I think giving a stone pick would be a waste of one slot

ClobberXD: Tools like pickaxes are the reason why those slots are considered to be prime real-estate. A stone pickaxe certainly wouldn't be a waste of one slot, IMO.

ClobberXD:
:+1: for 40 of each.

JostP:
Added stone pick back

PR #99, opened 2020-06-24
This is my PR.....

KodaTheGuy:


samuelstegerKOLLEGAH:
Map doesnt load completely for me just the base and this bugs the whole time. 
Could u maybe send me a zip file of the map ?

Zipitque:
How do you play it?

JostP:
Map doesn't work, try to fix it and test in CTF game as singleplayer.

samuelstegerKOLLEGAH:
From what I see there are broken barrriers, not all nodes are working, the map is really plane and really boring... It hasnt to do anything with twillight or vampires or whatever...

JostP:
This map is too much plain... We already have too many plain kind of maps in CTF. Don't ask how many plain maps we have rejected... I don't even remember all of them :thinking:

MinetestSam:
The map is broken and quite flat to be honest. Seems like you have placed the barrier nodes manually. Closing.

PR #101, opened 2020-06-28
The BOSS Fight

olliy1or:


Zipitque:
Nice Unique
Does not look like a copy of any map (Just a lil bit looking like isles)
9/10 rating
-Zipitque (The-Quack)

MinetestSam:
Hello, you need to upload a screenshot which shows the whole map and name it screenshot.png

JostP:
Re-open PR when Samuel is back and fix his map

PR #108, opened 2020-08-06
GreenLand Map

AnyBlue:
Hi! This is map for ctf. I hope you like it. Name: GreenLand


MinetestSam:
Hello, seems like you have uploaded two different maps in the same PR. I would suggest making separate PR for separate maps.

AnyBlue:
> Hello, seems like you have uploaded two different maps in the same PR. I would suggest making separate PR for separate maps.

Oh, yes. I wanted to add PR but I added to here. Idk why

AnyBlue:
> Hello, seems like you have uploaded two different maps in the same PR. I would suggest making separate PR for separate maps.

Hmmm... I can't delete map.mts

PR #111, opened 2020-08-12
Pyramids of Akmanrah

Alone0ne:


MinetestSam:
You need to fix the names of the files. For example, the file containing the schematic should be map.mts. Likewise, screenshot containing file should be screenshot.png and config file should be map.conf

Alone0ne:
just extract the zip file,you will find all the required files (in names preferrred )in a folder,its just a zipped file

MinetestSam:
> just extract the zip file,you will find all the required files (in names preferrred )in a folder,its just a zipped file

Sorry but that isn't the job of the map reviewer.

JostP:
I wanted to test your map today, but since `.conf` file is invalid, I can't do it. Check other map's `.conf` files to find an issue in yours.

Also, please, before making a PR, test map as singleplayer in CTF game. PR must contain `map.mts`, `map.conf` and `screenshot.png`.

Screenshot should show whole map from air and not only one pyramid like your screenshot.

JostP:
continue in #167

PR #112, opened 2020-09-11
Create snipers playground

coolest-yapa:
[ctf_55.zip](https://github.com/MT-CTF/maps/files/5211715/ctf_55.zip)


snipers playground
```diff
@@ -0,0 +1 @@
+
```
coolest-yapa, line 1: screenshot_20200911_002325

PR #119, opened 2020-10-06
Snow Land

RyanZem:
New map for CTF


JostP:
I reviewed your map:

1. screenshot should be 2:3 ratio, for example 400x600px is the best option.
2. Remove chests as they are not supported in CTF game.
3. Mese blocks around base are conspicuous, they doesn't look very nice. Use other node. You can also use `base_node = ctf_map:<item>` to set which blocks will be in base.
4. Your map is very plain what makes it similar to other maps. I think it would be nice to add mine with some iron and mese without diamonds. You know that maps with better weapons are the most popular. Give each side suitable amount of iron and mese, for example 40-70 iron and 20-30 mese per each side. From what I see, map does not have any iron underground currently. Add hint like `hint = search for a mine on the map!`

Everything else looks good :+1: and I will approve your map as soon as these points are solved.

JostP:
Looks good, you just need to update screenshot :+1: .

JostP (approved):

snowland/map.conf
```diff
@@ -0,0 +1,14 @@
...
+team.2.color = blue
+author = Elvis26
+rotation = z
+h = 140
+team.2 = blue
+name = Snow Land
+initial_stuff = default:pick_steel,default:torch 10
+r = 115
+license = CC BY-SA 4.0
+hint = You can find some ores in the mine
```
Panquesito7, line 14:
```suggestion
hint = You can find some ores in the mine.
```

RyanZem: @Panqesito7 Done.

Panquesito7 (approved):
Didn't test, but all looks good. 👍

snowland/map.conf
RyanZem: > 

@Panquesito7 Done

PR #125, opened 2020-10-13
Add "cave_sea" map by sodomite

sodomite:


AnyBlue:
@sodomite You need a pickaxe in your and a normal license

cave_sea/map.conf
```diff
@@ -10,9 +10,9 @@ rotation = z
...
 name = Cave Sea
 base_node = ctf_map:glass
-initial_stuff = grenades:frag 3,default:mese_post_light 99
+initial_stuff = default:pick_stone,grenades:frag 3,default:mese_post_light 99
 --chests.1.n = 1000
 --chests.1.from = -230,-230,-230
 --chests.1.to = 230,230,230
 treasures = default:pick_diamond,1,0,1;shooter:ammo,1,2,999;shooter:shotgun,1,10,1;shooter:machine_gun,1,10,1;grenades:frag,1,10,999
-license = LOL
+license = MulanPSL2
```
Thomas--S, line 18: I don't think that this is a valid Free Software/Open Source license.

JostP:
Change draft to Pull request when you finish

cave_sea/map.conf
```diff
@@ -0,0 +1,18 @@
+team.2.color = blue
+team.1.pos = (-7,2,6)
+team.2 = blue
+team.1.color = red
+team.1 = red
+team.2.pos = (-7,2,-6)
+h = 150
+r = 75
+rotation = z
+author = sodomit
```
Panquesito7, line 10:
```suggestion
author = sodomite
```

sodomite: no, sodimit is my ingame nick

cave_sea/map.conf
```diff
@@ -0,0 +1,18 @@
...
+name = Cave Sea
+base_node = ctf_map:glass
+initial_stuff = grenades:frag 1,default:mese_post_light 99
```
Avyukt1, line 13: I don't think that this will be useful
```suggestion
initial_stuff = grenades:frag 1,default:pick_mese,default:torch 10
```

Panquesito7: Maybe 20 or 20+ torches will be needed in case it's a cave.

JostP: default torches doesn't work in water.

Avyukt1: Yes, I have made those changes but I don't think that we need too many mese postlights.

cave_sea/map.conf
```diff
@@ -0,0 +1,18 @@
...
+team.1.color = red
+team.1 = red
+team.2.pos = (-7,2,-6)
+h = 150
+r = 75
+rotation = z
+author = sodomit
+name = Cave Sea
+base_node = ctf_map:glass
+initial_stuff = grenades:frag 1,default:mese_post_light 20
```
Panquesito7, line 13: Why mese post lights? Why not torches?

Avyukt1: Torches don't work in water.

cave_sea/map.conf
sodomite: Why do you want to add mese pick to initial stuff?

sodomite:
It is still unfinished, but I think it is playable now.

Panquesito7:
> It is still unfinished

IMO if it's still unfinished it should be still a draft PR.

sodomite:
> > It is still unfinished
> 
> IMO if it's still unfinished it should be still a draft PR.

I removed signs, now it is finished. I will add some more things and probably make it look better.

HobbitPower (commented):
Imo 999 nades, ammos etc. are way to high for a ctf map per chest. Even if all chests combined contain this amount it would be to high. And then we got a lot of shotguns, smgs etc. too.

cave_sea/map.conf
```diff
@@ -0,0 +1,18 @@
...
+chests.1.n = 300
+chests.1.from = -75,-75,-75
+chests.1.to = 75,75,75
+treasures = default:pick_diamond,1,0,10;shooter:ammo,1,2,999;shooter:shotgun,1,10,1;shooter:machine_gun,1,10,1;grenades:frag,1,10,999;sniper_rifles:rifle_magnum,1,10,1
```
JostP, line 17:
```suggestion
treasures = default:pick_diamond,1,0,10;shooter:ammo,1,2,5;shooter:shotgun,1,10,1;shooter:machine_gun,1,10,1;grenades:frag,1,10;sniper_rifles:rifle_762,1,10,1
```

I think shotguns and smgs could be OP on this map, especially because there are a lot of tunnels so one can easily kill you.
Maybe adding some more items to custom treasure wouldn't be a bad idea.

cave_sea/map.conf
JostP, line 13: I don't think giving grenade as initial stuff is a good idea. Picking enemy's flag is quite hard there and giving grenade to everyone could make it even harder.

JostP:
Tested.
Very interesting and detailed map, but unfortunately it has some big issues:
- It is too hard to capture, hm actually let's say impossible. Players can't use grapple hook after they have picked flag. There is only one small corridor from base to water.
- infinite grenades, smgs, shotguns, magnums... make it even harder to end match
- using grapple hook to move into secret corridors doesn't seem like a really good idea: on some places it is really hard to hit right place with hook and grapple hook gun can run out after some shots.

Some other players have tested it as well and they have pointed out the same issues. It cannot be added in current state, but you can of course try to improve it, but it should be changed a lot. As well, you might want to use ladders instead of apples.

DiamondPlane:
What about knights? They just can't use grapple hooks, even when they don't have the flag.

RyanZem:
In CTFv3 sharp shooters can only use ladders, they can't use grappling hooks. The map look is like a maze and it's difficult to capture the flag as JosP said.

PR #131, opened 2020-10-19
Add "diamond_hunt"

JostP:
(This pull request was generated by a script)


winniepee:
need a screenshot

JostP:
This map is **very** similar to Tunnel map. We accept only unique maps.
Also team have more good ores than other (15 mese block and 25 diamonds vs 12 mese blocks and 19 diamonds.

PR #135, opened 2020-10-24
Remove "Swamplands"

PiMonkeyServers:
Removal of the map Swamplands by popular demand and at the request of the creator


JostP (approved):
Looks good

PR #137, opened 2020-10-25
Add new "A Kingdom in the Nether" revisions, by Waterbug

KaylebJay:
Have not tested - this is just copypasta from Waterbug, who made the map.


JostP:
It doesn't look like you included correct files. If only schem changes, how is possible that bases are closer each other?

PR #140, opened 2020-10-28
Big Stone Mushroom And Nature by DiamondPlane

AnyBlue:
Map for CTF


KodaTheGuy:
**map.conf looks great!**

PR #153, opened 2020-11-19
Two Volcanoes Map by Oliver4

AnyBlue:
New map by Oliver4


olliy1or:
The map has no license

AnyBlue:
> The map has no license

I will ask him.

volcanoes/map.conf
```diff
@@ -0,0 +1,13 @@
...
+team.2.pos = (15,-63,-97)
+team.1.pos = (16,-63,101)
+team.1.color = red
+team.1 = red
+author = Oliver4
+rotation = z
+h = 140
+team.2 = blue
+name = Two Volcanoes
+initial_stuff = default:pick_steel,default:torch 10,shooter_guns:pistol
```
Avyukt1, line 11: This gives an unloaded pistol,
```suggestion
initial_stuff = default:pick_steel,default:torch 10,shooter_guns:pistol_loaded
```

Avyukt1 (changes requested):

Avyukt1:
This map doesn't have bugs. Also It has a good treasure. But the land flat and I think you should add more trees or mountains in the free space.

olliy1or:
imo the map is a little plain 
cant use the volcanoes for anything much
maybe add a little more builds to the sides and near the bases?

JostP:
Map works perfectly, but unfortunately it map won't be added:
- very plain, it isn't unique , we already have to many of them.
- we already have to many maps with good ores such as mese and diamond.

PR #154, opened 2020-11-27
Green Land Map by DiamondPlane

AnyBlue:
Green Land is back! This is my first map, by the way and I edited it a bit.


Avyukt1 (changes requested):

JostP:
I don't think this map will be accepted:
- It is not unique (very similar to better mini_village)
- Very plain, we already have too many of them
- A lot of mese ore, players are complaining that we already have too many mese and diamond maps.

AnyBlue:
> I don't think this map will be accepted:
> 
> * It is not unique (very similar to better mini_village)
> * Very plain, we already have too many of them
> * A lot of mese ore, players are complaining that we already have too many mese and diamond maps.

I didn't know that you don't want mese.

PR #156, opened 2020-11-27
"The Different Biomes" by winniepee0

winniepee:


AnyBlue:
Looks nice. But I recommend to change aspen on pine because in Minetest no aspen on snow...

olliy1or:
This PR is making changes to another map, submitting a map should only add the maps .mts .conf and a screenshot nothing else

abandoned_isles/map.conf
```diff
@@ -8,6 +8,7 @@ rotation = x
 h = 140
 team.2 = blue
 name = Abandoned Isles
+hint = Find a bunch of mese blocks!
```
Panquesito7, line 11: Please revert this change.

map.conf
```diff
@@ -0,0 +1,14 @@
+team.1.pos = (87,-26,100)
```
Panquesito7, line 1: Please remove this file. There's a `map.conf` for Different Biomes map already.

JostP:
You should create new branch from master and then make a PR.

winniepee:
Ok

PR #158, opened 2020-11-29
Add Snow and Lava map by realyg

AnyBlue:
A map made by realyg, Amine35


winniepee:
Seems okay, no bugs found.

Avyukt1:
Can you name the map like
`Snow vs Lava
or
`Water vs Lava`
?

olliy1or:
The map is a bit like lava puddles map
gives a big advantage to reds because the whole snow side is made out of snow and snow can be easily broken by guns

JostP:
I agree with olliy.
- very plain
- there are already maps like this on server
- something quickly WorldEdited

I also asked discord community and they agree too.

PR #164, opened 2020-12-05
Remove Big Ocean by gameit

Avyukt1:
This is a normal map, it also lags. Better to remove it as it is not good for PVP.


JostP:
Do not interfere in map reviewers work. If maps was bad, discord community vote would show that.

PR #167, opened 2020-12-07
Add new "Pyramids of Ahkmenrah" map by Darkf4antom and SoulSeeker

JostP:


ahkmenrah_pyramids/map.conf
```diff
@@ -0,0 +1,19 @@
+name = Pyramids of Ahkmenrah
+author = Darkf4antom and SoulSeeker
+hint = There is an invidible labyrinth under your base and a treasure at the end of it!
```
JostP, line 3:
```suggestion
hint = There is an invisible labyrinth under your base and a treasure at the end of it!
```

JostP (commented):
Looks good :+1:

PR #170, opened 2020-12-11
Lava Puddles Map Debug

Avyukt1:
There was a small bug in my map `Lava Puddles` And I have made changes and tested it. Its important to merge it before my map is updated in the server.


JostP:
What is that bug?

Avyukt1:
> What is that bug?

There is Indestructible Ladder on the brick wall for one base and nothing such thing on the other.

JostP:
> > What is that bug?
> 
> There is Indestructible Ladder on the brick wall for one base and no such thing on the other.

Ok will merge today.
Map is already on server btw. But this bug is not that big problem as you said.

JostP (approved):
Looks good :+1:

PR #173, opened 2020-12-13
Fix small bug in "Pyramids of Ahkmenrah" map

JostP:
- PR removes 2 invisible nodes per labyrinth so players don't need to jump anymore to solve the labyrinth,
- removes indestructible nodes in one of the tunnels,
- adds stone pick to inventory.


JostP (commented):
Looks good :+1:

PR #176, opened 2020-12-15
Changes to Mini Village map

AnyBlue:
Added mine, removed some houses, made the glass in village houses breakable, added bronze sphere, added hint, fixed chairs in a base house. I don't really remember what we added yet but I think it's everything what we made.


mini_village/map.conf
```diff
@@ -11,3 +11,4 @@ name = Mini Village
 initial_stuff = default:pick_steel,default:torch 5
 r = 115
 license = CC BY-SA 4.0
+hint = You can find treasures in the mine.
```
JostP, line 14:
```suggestion
```

JostP: Hint is not needed since there is no mese or diamonds in mine.

JostP: once you commit this, I will merge your map

JostP:
Nice update. Maybe you removed too many houses, but let's test how will players like the update.

JostP (approved):

Panquesito7:
> removed too many houses

Probably. I think they should add some back.

PR #177, opened 2020-12-16
Wall Of Wisdom by Fox/Whitzy

Fox3058:


AnyBlue:
Please rename the Screenshot on screenshot

wall_of_wisdom/map.conf
```diff
@@ -0,0 +1,13 @@
+team.1.pos = (-5,8,94)
+r = 115
+initial_stuff = Default:cobble 50,default:torch 20,default:pick_steel
```
Avyukt1, line 3:
```suggestion
initial_stuff = default:cobble 50,default:torch 20,default:pick_steel
```
You shouldn't use capital letters else that will not give the correct item.

Avyukt1 (changes requested):

Avyukt1:
This map is **too plain**. add more forest or other constructions.

This map is just like The Wall by Thomas-S

JostP:
re-open when you finish

PR #178, opened 2020-12-16
Add new map The Prison by winniepee0

AnyBlue:
New map by winniepee0


Avyukt1:
Did you add more constructions/forest in the remaining area?

MoreKat:
Very plain, flat map. The prison in the center of the map looks nice but doesn't add much to the map as you can just walk around it to get to the other team's base. Chest zones and map.conf seem fine.

JostP:
I can say that design of the prison is good, but unfortunately the rest part is very plain. I can tell you from my experience that players will just walk around it, as Kat said. From what I have seen, discord community don't want this map added either.
I would suggest including this prison in any other map with more buildings and diverse terrain as the prison design itself is quite good.

Avyukt1:
smugler5, realyg, winniepee0, Darkf4antom & SoulSeeker are making a map called `Bunker`. I think we should add the prison there because it also has quite flat land and the Prison will just Cover it.

JostP:
> smugler5, realyg, Tryhard_, winniepee0, Darkf4antom & SoulSeeker are making a map called `Bunker`. I think we should add the prison there because it also has quite flat land and the Prison will just Cover it.

Yes good idea to use it there.

PR #124, opened 2020-10-09
Add "plane_crash" map by minetevc

JostP:
(This pull request was generated by a script)

PR #185, opened 2020-12-30
Add new "The Legend of Capturing" map by gameit

JostP:


TSafa-23:
Looks great. Please merge soon!

TSafa-23 (approved):

JostP:
Needs better screenshot.

Avyukt1:
What's this?
image

Avyukt1:
@JostP 
The [plane_crash](https://github.com/MT-CTF/maps/pull/124) map has made some changes in this map's outer side. Maybe you should take a look.

Avyukt1:
image
image
image
 One base is inside a castle and has more protection than the other one which is on a bridge.
image
image

capture_legend/map.conf
```diff
@@ -0,0 +1,16 @@
+name = The Legend of Capturing
+author = gameit
+
+r = 87
```
Avyukt1, line 4:
```
r = 87
h = 108
```
These lines are creating error!

capture_legend/map.conf
```diff
@@ -0,0 +1,16 @@
...
+h = 108
```
Avyukt1, line 5: These lines are creating errors
```
r = 87
h = 108
```

JostP: I don't think so. I get no errors. You will get error if h isn't even number.

Avyukt1: See the screenshot
https://github.com/MT-CTF/maps/pull/185#issuecomment-752921597

Avyukt1: I don't even have the plane_crash in that folder anymore.

Avyukt1: image
This is outside the barrier.

Avyukt1: image

JostP:
I have to say that this map is really good :+1:.
But it still have some issues:
- door c. It would be great if it was removed with next update, but it is ok for now.
Zaslonska slika 2020-12-31 15-56-00
- Blue base is a lot more protected than red's. Like Avyukt1 said. I am sure that this will have to be updated in the near future as it is quite serious issue. But I think we can still add it and see what players say.

Zaslonska slika 2020-12-31 15-48-48

Zaslonska slika 2020-12-31 15-48-53

JostP:
I think custom chest zones would be a great addition.

capture_legend/map.conf
JostP: create new world and then try.

Avyukt1: Yes, new world is working fine.

JostP: this happened because you created new world with plane crash included.

Avyukt1 (commented):
Yes, This map will be 1 of the best ones in CTF.

PR #188, opened 2021-01-02
Add Compass map

Emojigit:
Also take from Karsthafen map


Emojigit:
WIP until https://github.com/MT-CTF/capturetheflag/issues/728 solve

JostP:
bases are 34 meters away from each other. I don't think this is a good idea. Try to place bases more far away, otherwise there will be the same problem as in currently disabled Castle Trouble map, every match was less than 5 min long. #172

Emojigit:
> bases are 34 meters away from each other. I don't think this is a good idea. Try to place bases more far away, otherwise there will be the same problem as in currently disabled Castle Trouble map, every match was less than 5 min long. #172

But the map is hard to find the way in the castle.

if you find way out of the castle, you need a long time to go.

JostP:
And if you use blocks to climb to the top of the castle? As medic you can easily finish map very quickly IMO. Build up, jump to enemy base, take flag, build up back to the top, go to your base.

JostP:
Feel free to re-open anytime when you fix the issues

PR #139, opened 2020-10-27
"Mulu Hills" by winniepee0

winniepee:

PR #190, opened 2021-01-04
Green Hills Map by DiamondPlane

Toby1710:
New map

PR #191, opened 2021-01-09
Add New Map: "Mulu Hills" by winniepee0

Avyukt1:
Adopted from https://github.com/MT-CTF/maps/pull/139


JostP:
This map is similar to Green hills #190 and I can say that Green hills is more interesting map since it has more buildings and features. I asked discord community and it seems that more players don't want it be added because we already have better Green Hills map.

My thought is that the hill is too high and players will have problems with climbing to the top what could cause boring gameplay.

PR #193, opened 2021-01-21
Add New Map "Beach"!

moreavy:


HobbitPower:
* Found 8 diamonds and 16 mese so far, a bit too much for me. But it is fairly divided between both sides:
ores
* remove blanks between items at initial_stuff
* There are nodes which can not be loaded. remove butterflies and firefly. the default chest will be probably found in the cobblestone dungeons
image
* I am not sure about the underground tunnel. The obsidian glass and the stone is breakable. this can cause some problems with the lava. If you did this on purpose: okay, discussable for me. if not: you can use indestructible glass (i know obsidian glass looks better) and indestructible stone:
image
* license is missing

moreavy:
> * Found 8 diamonds and 16 mese so far, a bit too much for me. But it is fairly divided between both sides:
>   ores
> * remove blanks between items at initial_stuff
> * There are nodes which can not be loaded. remove butterflies and firefly. the default chest will be probably found in the cobblestone dungeons
>   image
> * I am not sure about the underground tunnel. The obsidian glass and the stone is breakable. this can cause some problems with the lava. If you did this on purpose: okay, discussable for me. if not: you can use indestructible glass (i know obsidian glass looks better) and indestructible stone:
>   image
> * license is missing

Moomoodude made his/her map first time, I am guiding him on discord.

xenonca:
Screenshot would be nice too :)

xenonca:
> Found 8 diamonds and 16 mese so far, a bit too much for me. But it is fairly divided between both sides:

@HobbitPower Just makes 2 dia swords and 4 mese swords per team, I think that's ok :)

HobbitPower:
> > Found 8 diamonds and 16 mese so far, a bit too much for me. But it is fairly divided between both sides:
> 
> @HobbitPower Just makes 2 dia swords and 4 mese swords per team, I think that's ok :)

@xenonca hm yeah, i guess its okay :D just not a fan of to many diamond/mese swords ^^

JostP:
To be honest, map is very similar to some other maps. It is plain and it has some good ores (too many maps with mese and diamond is not good).

But to see it from bright side, I can say that it has no bugs like broken barriers :+1: . So you now know how to make a map. But note that only unique good maps will be added.

Zaslonska slika 2021-01-28 16-51-11

moreavy:
> P.S. You should choose license for your maps.

What does this mean?

JostP:
> > P.S. You should choose license for your maps.
> 
> What does this mean?

Sorry, I thought there wasn't license, but I see it now.

beach/map.conf
```diff
@@ -0,0 +1,13 @@
...
+team.2.pos = (26,4,-99)
+team.2.color = blue
+author = MoomooDude
+rotation = z
+h = 140
+team.2 = blue
+name = Beach
+initial_stuff = default:cobble 40,default:torch 10,default:pick_steel
+r = 115
+license = CC BY 3.0
```
Panquesito7, line 13: Why CC BY 3.0? Why not CC BY 4.0? Creative Commons itself recommends to use the latest version of their licenses.

Panquesito7 (commented):
This map is also missing screenshot.

beach/map.conf
JostP: > Why CC BY 3.0? Why not CC BY 4.0? Creative Commons itself recommends to use the latest version of their licenses.

It is recommended to use latest version, but note that map maker can choose any open source license.

Issue #157, opened 2020-11-28
Middle Barrier Alignment


xenonca:
> You need to select Z=0 while placing the barriers.
> If you do X=0, you will get wrong initials of the bases and the barrier will be crashed.

_Originally posted by @Avyukt1 in https://github.com/MT-CTF/maps/issues/124#issuecomment-723125866_

I don't think that it matters if you place the middle barrier along the Z or the X axis, I mean that's why you have a choice. However I've tried to export a map too short time ago and the middle barrier wouldn't change direction, no matter which axis I selected.

Avyukt1:
Maybe you should add this issue in the [MT-CTF/capturetheflag](https://github.com/MT-CTF/capturetheflag) to make changes in the `ctf_map` mod.

savilli:
Fixed in V3

PR #194, opened 2021-01-23
Add 'Sam's Mansion'

xenonca:
Initially I wanted to have the middle barrier aligned to the other axis but due to a [bug](https://github.com/MT-CTF/maps/issues/157) in the map_maker mod I had to look for a solution. Instead of rebuilding this pretty detailed map I decided to slightly move the bases so that the middle barrier can still go between them.


MoreKat:
Very original and detailed map. I really like the secret pathways and the many levels. My only concern is that the bases might be a bit unbalanced, red has the high ground and is more tucked away in the corner. Another problem that might arise is that this map can be ended very quickly. During the build time both teams can go to the other team's base and immediately try to capture when the game starts. This could, on the other hand, also create some interesting matches and force players to try out a new play style.

xenonca:
> red has the high ground and is more tucked away in the corner  

I see the problem here yea... I think this has also a lot to do with the fact that I had to move the bases, else they'd be where I placed the ind_diamond_ore. Any suggestions where the bases would be located better?

> During the build time both teams can go to the other team's base  

This is because of the middle barrier alignment, else this wouldn't be the case. Unfortunately I don't see a good solution for this without having to redo the whole map.

sodomite (approved):
Good looking rats map.

moreavy:
Working Fine... No Bugs, also I liked it.

moreavy (approved):
Fantastic!

sams_mansion/map.conf
```diff
@@ -0,0 +1,27 @@
...
+team.2.pos = (46,-7,-5)
+h = 26
+r = 62
+rotation = z
+author = Xenon
+name = Sam's Mansion
+initial_stuff = default:pick_steel,default:desert_sandstone 50,default:torch 10,shooter:pistol
+base_node = ctf_map:wool_red
+hint = Don't miss the secret passages!
+license = CC BY-SA 4.0 | Skybox: CC_BY 3.0 by Jockum Skoglund (hipshot) and with adaptions from congusbongus (https://opengameart.org/content/interstellar-skybox-png)
```
Panquesito7, line 15:
```suggestion
license = CC BY-SA 4.0 | Skybox: CC BY 3.0 by Jockum Skoglund (hipshot) and with adaptions from congusbongus (https://opengameart.org/content/interstellar-skybox-png)
```

sams_mansion/map.conf
```diff
@@ -0,0 +1,27 @@
...
+team.2.color = blue
+team.1.color = red
+team.2 = blue
+team.2.pos = (46,-7,-5)
+h = 26
+r = 62
+rotation = z
+author = Xenon
+name = Sam's Mansion
+initial_stuff = default:pick_steel,default:desert_sandstone 50,default:torch 10,shooter:pistol
```
Panquesito7, line 12:
```suggestion
initial_stuff = default:pick_steel,default:desert_sandstone 50,default:torch 10,shooter_guns:pistol
```

moreavy: Its should be 
```suggestion
initial_stuff = default:pick_steel,default:desert_sandstone 50,default:torch 10,shooter_guns:pistol_loaded
```
```properties
initial_stuff = default:pick_steel,default:desert_sandstone 50,default:torch 10,shooter_guns:pistol_loaded
```
Else it will give an unloaded pisol.

JostP: `shooter:pistol` works fine. Don't change it.

moreavy:
Can the `One Approval` label work here?

Panquesito7:
> Can the One Approval label be added?

I think it's only for those who approve and have triage access and above (in other words: maintainers).

JostP:
WIP at the moment, Xenon is making some changes

JostP:
Tested, works and look really nice :+1: 
But, since map is rotated chest zones doesn't work anymore. Try to fix it, IIRC you said that you had updated chest zones, so you probably forgot to include it in the update.

sams_mansion/map.conf
```diff
@@ -0,0 +1,28 @@
...
+team.2.pos = (-5,-7,-47)
+h = 26
+r = 62
+rotation = x
+author = Xenon
+name = Sam's Mansion
+initial_stuff = default:pick_steel,default:desert_sandstone 50,default:torch 10,shooter:pistol
+base_node = ctf_map:wool_red
+hint = Don't miss the secret passages!
+license = CC BY-SA 4.0 | Skybox: CC BY 3.0 by Jockum Skoglund (hipshot) and with adaptions from congusbongus (https://opengameart.org/content/interstellar-skybox-png)
```
JostP, line 15:
```suggestion
license = CC BY-SA 4.0. | Skybox textures adapted from the skybox "Interstellar skybox" (from https://opengameart.org/content/interstellar-skybox) by Jockum Skoglund (hipshot) and with adaptions from congusbongus (https://opengameart.org/content/interstellar-skybox-png), originally licensed under CC BY 3.0.
```

moreavy (approved):

sams_mansion/map.conf
xenonca: I'm against these changes, why the extra point at the end of the License?
The other suggestion of skybox textures doesn't make sense. I'm not trying to re-license it, I'm just including their media in my PR and thus mentioning their license in it. The adaptions have been made by congusbongus, neither by Jockum nor by me so if I add your suggested changes it doesn't mean the same anymore.

JostP: I see your point, it really doesn't mean the same, yes.
It was just a suggestion since moon map's "skybox license text" was like suggested one. It was approved by ANAND and Thomas-S.
I mean, both versions are valid.

JostP (approved):
Chest zones work :+1:

MoreKat (approved):

PR #195, opened 2021-01-28
Remove Pine Forest and disable Towers and Lava Puddles

olliy1or:


moreavy:
Are these the map voting results?

moreavy (approved):
I always support the map voting results.

JostP:
> I always support the map voting results.

One of map voting result is that you will have to improve lava puddles map because it got average vote less than 2.7

moreavy:
> > I always support the map voting results.
> 
> One of map voting result is that you will have to improve lava puddles map because it got average vote less than 2.7

Ah okay. I am working on it. I'm planning to change its name to `Volcanic Battlefield`.

JostP:
Lava Puddles map will be disabled temporary until it is updated.

towers/map.conf
```diff
@@ -44,3 +44,5 @@ chests.7.to = (-100,-24,-1)
 chests.8.n = 10
 chests.8.from = (35,-24,-100)
 chests.8.to = (100,-24,-1)
+
+disabled = true
```
JostP, line 48: It would be better if you added `disabled = true` to first line

JostP: So when you open file, you immediately see that it is disabled

JostP (approved):

PR #199, opened 2021-02-01
Updates to Mini Village by Elvis26 and DiamondPlane

Toby1710:
We added jail, helicopter, fire truck, flaming house, small camp, river + bridge, tunnel from one house to another, added fountain in the yard, two small hills.
image


moreavy (approved):
working fine

JostP:
Nice update, works fine :+1: 
But There is a bug:
Zaslonska slika 2021-02-01 19-46-04
There is no barrier under water. To fix it, just select red barrier as region and do:
`//replace default:river_water_flowing ctf_map:ind_glass_red` and then `//replace default:river_water_source ctf_map:ind_glass_red`. Or you can also manually place red barrier.

One more thing:
It would be nice if all windows in base were break-able.
Zaslonska slika 2021-02-01 19-44-48
Currently windows in first floor are indestructible.

Toby1710:
> Nice update, works fine 
> But There is a bug:
> Zaslonska slika 2021-02-01 19-46-04
> There is no barrier under water. To fix it, just select red barrier as region and do:
> `//replace default:river_water_flowing ctf_map:ind_glass_red` and then `//replace default:river_water_source ctf_map:ind_glass_red`. Or you can also manually place red barrier. It is a bug because players can go under barrier before match starts.
> 
> One more thing:
> It would be nice if all windows in base were break-able.
> Zaslonska slika 2021-02-01 19-44-48
> Currently windows in first floor are indestructible.

Replacing doesn't work, we tested and it's the same.

JostP (approved):

PR #200, opened 2021-02-01
Add Ewok Village

Anthony-De:


moreavy:
P.s: @Anthony-De 
You need a license for your map.
Go to https://github.com/MT-CTF/capturetheflag/tree/master/mods/ctf/ctf_map/map_maker#license to learn more.

ewok_village/map.conf
```diff
@@ -0,0 +1,13 @@
+team.1.pos = (38,-10,39)
+r = 50
+initial_stuff = default:pick_steel, default:junglewood 50
```
Toby1710, line 3:
```suggestion
initial_stuff = default:pick_steel,default:junglewood 50,default:torch 3
```
Don't need to put space. Also add torches

Toby1710:
I suggest to add something yet, it looks plain. But: I didn't see any erros, it doesn't look for me similar to other maps, it looks nice.

Toby1710 (commented):
I suggest to add something yet, it looks plain. But: I didn't see any erros, it doesn't look for me similar to other maps, it looks nice.

moreavy (commented):
It seems good from the screenshot that its a new idea for map design. It looks good and will probably test it today.

moreavy:
@Anthony-De Map has bugs. The bases are outside the map.
image

MoreKat:
It also seems like the map doesn't have a ceiling?

MoreKat:
It looks like the issues have been fixed. I really like the map, very original and I think it could be a good replacement for the disabled 'Towers' map. However, I think the map could use some more details, a few suggestions:
- some terrain variation (lakes, hills)
- make the small trees partly hollow with houses (for the Ewoks) in the hollow parts
- possibly some branches? That might be hard to build though.

ewok_village/map.conf
```diff
@@ -0,0 +1,14 @@
...
+team.2.color = blue
+team.2.pos = (-38,-19,-39)
+r = 50
+rotation = z
+h = 120
+author = AnthonyDe and JediKnight
+name = Ewok Map
+base_node = ctf_map:junglewood
+initial_stuff = default:pick_steel,default:junglewood 50,default:torch 3
+license = CC BY-SA 4.0
```
Panquesito7, line 14: Missing end line.
```suggestion
license = CC BY-SA 4.0

```

ewok_village/map.conf
```diff
@@ -0,0 +1,14 @@
...
+team.1.color = red
+team.1.pos = (38,-19,39)
+team.2 = blue
+team.2.color = blue
+team.2.pos = (-38,-19,-39)
+r = 50
+rotation = z
+h = 120
+author = AnthonyDe and JediKnight
+name = Ewok Map
```
Panquesito7, line 11: Shouldn't here be `Ewok Village`?
```suggestion
name = Ewok Village
```

HobbitPower:
First of all, i like the seize of the map and like Kat said, it seems like a good replacement for towers.

*i got some errors while joining. someone else got them too? 
image

* for me there are too many ores, but thats a matter of taste. (i am not a fan of too many ores in general ^^)
image

otherwise i like the map. ofc you could add more if you want but you dont have to.

moreavy:
Does that map have mese?

moreavy:
Map is cool, but I just don't think the map needs so much mese and diamond, I recommended add good treasures like shotgun and grenades but remove mese and diamond.

ewok_village/map.conf
```diff
@@ -0,0 +1,14 @@
...
+team.2 = blue
+team.2.color = blue
+team.2.pos = (-38,-19,-39)
+r = 50
+rotation = z
+h = 120
+author = AnthonyDe and JediKnight
+name = Ewok Village
+base_node = ctf_map:junglewood
+initial_stuff = default:pick_steel,default:junglewood 50,default:torch 3
```
moreavy, line 13:
```suggestion
initial_stuff = default:pick_steel,default:junglewood 50,default:torch 3
treasures = shooter:shotgun,0.3,5,1,10;grenades:frag,0.5,7,2;default:torch 3,4,8,1;default:stone 50,3,9,2;default:junglewood 50,1,8,0.7;shooter:pistol,2,8,4;grenades:smoke,1,8.9,2;shooter:ammo,5,9,7
```

Anthony-De:
For an ore comparison, I dug out less than half of one side on "The Wall".
image

moreavy:
> For an ore comparison, I dug out less than half of one side on "The Wall".
> image

We already have lots of maps with mese and diamond and I don't think we need more. Also, your map is fun without mese and diamond and I don't feel the need.

MoreKat (commented):
What I'm worried about in this map is that each base is really tucked away in the corner. Add to that the **mese and diamond swords** (which shouldn't be removed imo, maps with mese and diamonds are generally more fun and the existing maps with those ores are not similar to this one) and the fact that the base is **in the sky, with only a few narrow pathways** leading to it, and you get a map in which it's quite hard to capture.

This could be avoided by maybe moving the bases more to the middle of each side. Some things that I'm still missing are some more terrain variation (doesn't have to be much, just to give a player the possibility to take cover) and branches (those could be nice as they could be a good start for a skybridge). This is just a suggestion though, I really like the map as it already is now.

moreavy:
Yes, it will be better if the positions of the bases changes.

ewok_village/map.conf
```diff
@@ -0,0 +1,15 @@
...
+license = CC BY-SA 4.0
+treasures = default:ladder,0.3,2,10,20;default:junglewood,0.3,5,30,60;default:axe_steel,0.15,2,1,2;shooter:shotgun,0.05,2,1;grenades:frag,0.1,2,1;grenades:smoke,0.1,2,1;shooter:machine_gun,0,03,2,1;shooter:pistol,0.4,2,1,4;shooter:ammo,0.3,2,1,10;medkits:medkit,0.8,5,2
```
JostP, line 15: So you removed:
- all blocks from chests
- sniper rifles
- shovels
- picks
Everything on purpose?

moreavy: Yeah, sniper rifles should be in the `initial_stuff`

JostP:
Tested:
- no bugs found :+1: 
- suitable map size, bases are only 100 far from each other :+1: 
- unique
-------------------------
- What I am worried about is, as Kat already said, that it is hard to capture. Moving bases to lower position is a bad idea as player would use only ground to fight. So the best solution is probably to move the bases closer together and/or make more pathways leading to the middle and to the ground.
- What I miss are some treetops with green leaves since that is Ewok Village. Well it is just an aesthetics but it would definitely makes map nicer and more attractive. Even a few small green treetops/green leaves on the bridges would make it better.
- Also check Kat's good review

In general the map looks promising and will hopefully be merged after next map voting if you try to fix issues listed above.
P.S. Map with forest skybox could look great.

Difference:
slika
Zaslonska slika 2021-04-11 17-25-58

ewok_village/map.conf
Anthony-De: @moreavy It is as I have intended.

HobbitPower (approved):
Found no big bugs so far.

Good choice with the chests at the tree crown. 

Some parts of the middle barrier are missing, but i dont know if thats a problem: 
image

Otherwise everything worked fine for me.  I like it 👍

ewok_village/map.conf
```diff
@@ -0,0 +1,15 @@
+team.1 = red
+team.1.color = red
+team.1.pos = (38,-19,39)
+team.2 = blue
+team.2.color = blue
+team.2.pos = (-38,-19,-39)
+r = 50
+rotation = z
+h = 120
+author = AnthonyDe & Kat & PadrePio
```
Panquesito7, line 10:
```suggestion
author = AnthonyDe, Kat, and PadrePio
```

MoreKat (approved):
I'm still a bit concerned about how hard it will be to capture, but I think we'll only know after the map has been played a few times. If needed, changes can still be made.

Panquesito7 (approved):

olliy1or (approved):

PR #202, opened 2021-02-07
Revive Wool map

JostP:
This PR removes old "Wool Land" map and adds new version. Some special features:
- Map is made of wool islands with a kill nodes under them. When you fall down, you definitely die.
- There is a castle exactly at the middle, so both teams meet and fight there. One team can occupy and easily defend it after that. It is quite hard to go around castle.
- 4 custom chest zones, two of them are only accessible after match has started. Moreover there is small iron mine accessible after match has started too.
- Map is relatively small and have more than one path to the enemy base, so matches shouldn't last long.
- Custom treasures
- Three sniper towers across the map.
- Set jump to 1.25
- Skybox

https://github.com/MT-CTF/capturetheflag/pull/783 Should be merged first.

Zaslonska slika 2021-02-07 12-10-08


JostP:
Walking on wool now makes sound.

JostP:
screenshot

MoreKat (commented):
Thanks for implementing my suggestion, looks awesome. Very unique, not only because almost everything is made out of wool, but also the map itself, something that the previous wool map lacked imo. I think this map is ready to be merged :)

PR #206, opened 2021-02-13
Moon: increase gravity a little

JostP:
Changes gravity due to reports of players that 0.16 gravity is a bit annoying.

### DIFF
--- a/moon/map.conf
+++ b/moon/map.conf
@@ -17,5 +17,5 @@ phys_speed = 1.25
 phys_jump = 1
 
 base_node = ctf_map:gravel
-initial_stuff = default:shovel_steel,shooter:pistol
+initial_stuff = default:shovel_steel,shooter:pistol,default:torch 3
 license = CC BY-SA 4.0. Skybox textures adapted from the skybox "Ulukai's space skyboxes" (from https://opengameart.org/content/ulukais-space-skyboxes) by Ulukai, originally licensed under CC BY 3.0. Earth with CC0 license (from https://purepng.com/photo/13214/nature-earth) was added to skybox_4.png. Rockets, machines, satellite dish, tunnels and ice were added by gimp and caoutiouspotato.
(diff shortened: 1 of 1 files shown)


Panquesito7 (commented):
I've tested it; to be honest, it isn't as fun as it was before. Maybe decrease a little bit the gravity?

moon/map.conf
```diff
@@ -12,9 +12,9 @@ team.2 = blue
...
 
-phys_gravity = 0.16
-phys_speed = 1.2
-phys_jump = 0.8
+phys_gravity = 0.3
+phys_speed = 1.25
+phys_jump = 1
 
 base_node = ctf_map:gravel
 initial_stuff = default:shovel_steel,shooter:pistol
```
moreavy, line 20:
```suggestion
initial_stuff = default:shovel_steel,shooter:pistol,default:torch 3
```

moreavy (commented):
Maybe adding torches in `initial_stuff` would be a good idea though.

moreavy:
I suggest the `decision needed ⚖️` label can be used here as @Panquesito7 and @moreavy have tested and it works fine. But what we need is just a final decision right now or maybe a change the the moon's `map.conf` file.

JostP:
> I've tested it; to be honest, it isn't as fun as it was before. Maybe decrease a little bit the gravity?

Thanks for testing. Yes, we haven't decided if this will be merged, that's why we need many people to tell their opinion.

olliy1or (approved):

moreavy (changes requested):
Not as fun as the old moon. Less Gravity makes the moon realistic. I didn't like the increased gravity of the moon.

MoreKat (commented):
Tested it and I wouldn't change it. The reduced gravity is what makes the map unique, and I don't think this would have a significant impact on people's opinion about the map.

Toby1710 (approved):

Panquesito7:
> Maybe adding torches in `initial_stuff` would be a good idea though.

This should be made in another PR.

JostP:
Will deal with this later.
I have some plans to improve moon map one day, currently it has some issues, on of them long matches.

PR #207, opened 2021-02-14
Add "Coral Reef" by -lipop-

moreavy:
New Map made by -lipop-


coral_reef/map.conf
```diff
@@ -0,0 +1,13 @@
...
+team.2.pos = (0,1,-74)
+team.1.pos = (0,1,76)
+team.1.color = red
+team.1 = red
+author = -lipop-
+rotation = z
+h = 140
+team.2 = blue
+name = Coral Reef 
+initial_stuff = default:cobble 99
```
Panquesito7, line 11: You should probably add 3 torches or more.

Panquesito7, line 11: Please include a pickaxe (at least a stone pickaxe).

Panquesito7 (commented):
Tested; no bugs found. But, there's too much water, players won't be able to PvP properly, the map is practically all water. Maybe add somewhere players can stand and PvP a bit (not sure if it's a good idea though).

Panquesito7 (commented):
From the screenshot, it looks cool. 👍

moreavy (commented):
When I tested, I didn't find any bug. But I found other water without the constructions very normal leaving the cliff over there. The map can be done better if the author adds more constructions.

JostP:
Still WIP or already finished?

moreavy:
> Still WIP or already finished?

Its finished according to -lipop-

coral_reef/map.conf
```diff
@@ -0,0 +1,13 @@
...
+team.1.pos = (0,1,75)
+team.2 = blue
+team.1.color = red
+team.1 = red
+team.2.pos = (0,1,-75)
+h = 140
+r = 115
+rotation = z
+author = -lipop-
+initial_stuff = default:pick_stone,default:torch 3,default:cobble 99,
```
Panquesito7, line 11:
```suggestion
initial_stuff = default:pick_stone,default:torch 3,default:cobble 99
```

Panquesito7 (approved):
Tested, no bugs found. The map looks cool; players will be able to PvP easier than before.
The only thing is the color of the wool in the blue base, it should be blue, not red (not sure if this is a good idea).

moreavy:
-lipop- made changes, I just need to update the screenshot.

RyanZem (commented):
I didn't review the map yet but I can see that you need to add yet some buildings or bridges.
This map looks nice, the bridge is nice too, but I can see that snipers can shoot in base. 
You can add a wall near the bases and snipers won't spawn-kill.

moreavy:
@RyanZem Done!

moreavy:
This map is flat and similar to Thomas-S's map, -lipop- told me to close this.

HobbitPower (commented):
Hm, i am torn between approval and refusal. 
Nice buildings (like the corals, base etc) but it looks a bit "thrown together". Also all fights will mostly take place at the middle glass path. 
Quite similiar to the Big Ocean map (although Big Ocean even has skybridges/walks).
I like the positions of the chests.

DiamondPlane (commented):
I think this map has too much unbreakable things... Almost everything that I wanted to dig is just unbreakable. It's hard to capture, I see... I think you should make at least the glass or wool around base breakable, plus I think it's similar to Big Ocean, like HobbitPower said. Also, you should update the screenshot. There are many things that are not on the screenshot.
But this map looks good for me, buildings are cool.
This is my review.

MoreKat (changes requested):
First of all, too similar to Big Ocean; an ocean with a bridge going from one base to the other and some random buildings on the sides. Something that I also notice is that the map lacks coherence: looks like a bunch of random, individual buildings were put together without respecting a general theme. For example, on one side of the map you have these realistic looking boats and on the other side cartoon sea creatures. The seemingly wrong proportions also add to this impression, e.g. the dolphin that's bigger than the boat. 

Another issue that I see is that capturing might be hard. Base is surrounded by shallow water, in which you can see an enemy coming from far away. Especially blue is at a disadvantage because there are cliffs next to their base, connected to the skybridge, which red can use to capture easily. There is no such thing on red's side.

PR #211, opened 2021-02-23
Map To Review: Hill CTF by Darkf4antom and SoulSeeker

moreavy:


moreavy:
I tested, here are my thoughts:

- Your map is too plain leaving the ugly hill in the middle
- The hill in the middle is not looking like a hill
- Nothing too much unique
- The ugly and broken construction surrounding the base will make it difficult to build reinforcements outside the base. I know players can make larges reinforcements but that will take more resources to build it.

image

Alone0ne:
WIP

RyanZem:
This map is too flat and I can't dig anything only stone.
Make this map more interesting and I can see beyond the glass barrier other dirt, replace it on the dirt what you made the map.

moreavy:
> This map is too flat and I can't dig anything only stone.
> Make this map more interesting and I can see beyond the glass barrier other dirt, replace it on the dirt what you made the map.

You should wait to test WIP maps.

Realyg-Hmar:
Can't say anything im totally agree with moreavy .

PR #212, opened 2021-03-14
Add cobble to Green Hills

DiamondPlane:


moreavy (commented):
I saw that Green Hills map doesn't need any cobble in initial stuff,
- The Map doesn't have tall buildings or towers
- Medics already get cobble when the (re)spawn. The medics would have too much blocks if they get extra in initial stuff.
- The quantity of cobble is huge for `initial_stuff`
- There already is a **STEEL PICKAXE** is `initial_stuff`, you can mine blocks and collect them.

DiamondPlane:
I just made it because Elvis26 suggested, many people liked it.

JostP:
I don't like idea to give every player 70 cobble. People will make pillars even more. If you want to add something for climbing then I suggest to add some ladders to initial stuff.

JostP (approved):

PR #213, opened 2021-03-29
Add "Sewer Under Village" by The Map-Makers Inc.

winniepee:


moreavy:
Not Tested, but according to the screenshot posted in Discord, This map is really similar to Plain Battle and Mini Village. Also was that the sewer underground is similar to the idea of the Bunkers Map.

PR #214, opened 2021-03-29
Add "Sewer Under Village"

winniepee:


moreavy:
Not Tested, but according to the screenshot posted in Discord, This map is really similar to Plain Battle and Mini Village. Also was that the sewer underground is similar to the idea of the Mayan Legend.

And also remember:
> A sewer is where all the wastes from toilets go.

moreavy:
Not Tested, but according to the screenshot posted in Discord, This map is really similar to Plain Battle and Mini Village. Also was that the sewer underground is similar to the idea of the Mayan Legend. Also seems to big and flat.

And also remember:
> A sewer is where all the wastes from toilets go.

sewer-village/map.conf
```diff
@@ -0,0 +1,22 @@
+team.2.color = blue
+team.2.pos = (-76,-54,-68)
+team.1.pos = (76,-54,68)
+team.1.color = red
+author = Map-Makers Inc.
+rotation = z
+h = 140
+team.1 = red
+team.2 = blue
+name = Sewer Under Village
```
moreavy, line 10:
```suggestion
name = Underground Sewer
```

HobbitPower (commented):
I already wrote some comments on discord (gameit is already fixing the stairs at the blue side + the 2 unbreakable blocks of the secret path). Have to agree, the map got some similarities to other maps like plain battle, mini village and bunkers. but it looks good, pretty buildings (especially the construct in the middle).
the map is big and some players maybe dont like this, but there are plenty ways to capture the flag + enhanced speed. so i am okay with it. 
could be interesting. Some players maybe wont like it (especially the underground part), but i like the combination of the underground and the surface. --> although i am still thinking that blue got a little advantage (very close to the secret path/ladder way through the house, red got a longer way).
i like the fact, that the mese is not so easy to reach.

moreavy:
> i like the fact, that the mese is not so easy to reach.

Mese is just in front of the base.

HobbitPower:
> > i like the fact, that the mese is not so easy to reach.
> 
> Mese is just in front of the base.

no there are 10 mese in general. look into the sky ;)

Realyg-Hmar:
> no there are 10 mese in general. look into the sky ;)

Yeah there mese and some diamond in the Ballons

sewer-village/map.conf
```diff
@@ -0,0 +1,22 @@
...
+r = 110
+initial_stuff = shooter:pistol,default:pick_steel,default:mossycobble 50,default:torch 10
+base_node = ctf_map:meselamp
+others = This map is made by realyg, liverpool, winniepee0, DiamondPlane, Elvis26, j45, and gameit.
+license = CC BY-SA 4.0
+phys_speed = 2
```
JostP, line 16:
```suggestion
phys_speed = 1.5
```

sewer-village/map.conf
```diff
@@ -0,0 +1,22 @@
+team.2.color = blue
+team.2.pos = (-76,-54,-68)
+team.1.pos = (76,-54,68)
```
JostP, line 3:
```suggestion
team.1.pos = (55,-21,72)
```

sewer-village/map.conf
```diff
@@ -0,0 +1,22 @@
...
+author = Map-Makers Inc.
+rotation = z
+h = 140
+team.1 = red
+team.2 = blue
+name = Sewer Under Village
+r = 110
+initial_stuff = shooter:pistol,default:pick_steel,default:mossycobble 50,default:torch 10
+base_node = ctf_map:meselamp
+others = This map is made by realyg, liverpool, winniepee0, DiamondPlane, Elvis26, j45, and gameit.
```
JostP, line 14:
```suggestion
others = This map was made by realyg, liverpool, winniepee0, DiamondPlane, Elvis26, j45, and gameit.
```

sewer-village/map.conf
JostP: This moves red base to the surface.

sewer-village/map.conf
JostP: 2 is too much unfortunately. 1.5 is better but still a lot.

JostP:
Tested. Works, no bugs found :+1:
Map is similar to some other maps, but I think it looks enough interesting and unique to be added. 
- It would be a lot better if one base was on the surface. Currently you have both bases underground, which means that players will rarely come to the surface. I [suggest](https://github.com/MT-CTF/maps/pull/214#discussion_r631861390) to move red base to the surface. This will make players use both "sewer" and village.
- Since Mini Village map was (temporarily) removed, it would be nice to have this map on the server. It looks like upgraded Mini Village map.
- 5 mese + 1 dia per side :+1: 
- Maybe map should have more passages which connects surface and "sewer" because red base will be on the surface. Let me know if you want to add more. It is not really necessary to do it tho.
- You should use better screenshot. Here you have one better:
screenshot

Realyg-Hmar:
I said to DP " the base of red/blue should be in the surface?" but they said "no" .
Ik that is good idea to make base on the top but they won't... im agree with u -sniper-.

winniepee:
Dang! haven't see this for a while. Let me change some stuffs ...

PR #219, opened 2021-04-18
Add Space Map by -lipop-

moreavy:
image


moreavy:
Cool Map with nice skyboxes and ores and with proper amount of gravity and speed, I am here itself approving it because Authors can't approve their Pull Request manually. As a Map Tester, I tested it and opened my opinion there.

DiamondPlane:
So, I tested it, and it seems to me that it's too hard to capture, plus CTF already have a map with space...
I understand that it's difficult to make a map like this but that's just my comment.

olliy1or:
This PR seem to be editing a lot more files than it probably should edit

PR #223, opened 2021-06-18
New map: Knight's Plain

W3RQ01:
I made this plain map for CTF and i'd like if it can be enabled into the server.


ksandronline:
Hi. im do small test. 

screenshot_20210618_163451

3 Moments: 
- need remove this blocks from map 
   default:chest
   tnt:tnt
- add screenshot in size 3:2 
- in game CTF with map not work.

W3RQ01:
ok give me 10 minutes

moreavy:
The map seems to be too plain according to @ksandronline's screenshot. Flat and plain maps are too boring these days...

W3RQ01:
it's completely plain apart for some caves... 
It's my first map and i was wanting to make a map that can have battles like The Wall...
There is also a secret passage that connects the two bases just need to find it :)

Realyg-Hmar:
This map is too plain .. u also need screen shot as @ksandronline said.

DiamondPlane:
Screenshot is missing

HobbitPower:
@W3RQ01 Do you plan to revise the map in the near future? The map would need to be updated to CTF v3 with the integrated map editor anyway. 
In general, the map is too flat and too reminiscent of already existing maps. Also, due to the position of the bases, you will mainly only play on one side of the map. A screenshot is also missing. So let us know if you still plan to change something about the map and/or update/convert it to v3.

HobbitPower:
The owner has not responded to any messages here or on Discord, so I am closing the PR.

PR #224, opened 2021-06-22
Add "The Olympics"

winniepee:
Add "The Olympics" by realyg, winniepee0, gameit and GabrielOMG


Realyg-Hmar:
The best map (:

DiamondPlane:
Didn't test it but on the screenshot it looks nice.

ksandronline:
Im inspected map. Very beatiful map. i want play this map.
we need to change it: default:mese_post_light_aspen_wood
screenshot_20210623_183336

and im think red base more shots are fired:

screenshot_20210623_183737

screenshot_20210623_183803

вut I have never been for sameness

MoreKat (commented):
Very similar to existing maps when you look at the structures, the environment itself. Only the theme is actually original.
Another concern of mine are the bases, they seem unbalanced to me, with the blue team being in favour. Their base is positioned higher, there's less cover around it, so they can see enemies coming easily, **and** it's very close to the map border, so they don't even have to watch their back. 
Red's base on the other hand is lower, with four towers around it (not sure what the function of those is), which only limit their view. There's also plenty of cover for enemies around it (two hills with a building on) and there's not much space between the base and the border, but still more than on blue side.

HobbitPower:
Closing this PR, new one can be found at #278

PR #228, opened 2021-08-14
New map "Pirate's Hideout"

olliy1or:


Panquesito7 (commented):
Maybe add a closer image of the map? It looks good, but I can't see it that good.

MoreKat (commented):
Very nice, detailed, unique map. However I wonder how hard it will be to capture, given that the base is completely surrounded by water, and there's not much possibility to tunnel. I suppose we will find out when the map has been added, and if it turns out to be quite hard to capture, you could add some more islands or so.

MoreKat (approved):

PR #230, opened 2021-08-25
adding simple_village map

farooqkz:
This is a simple village map without almost any modification. Generated using https://content.minetest.net/packages/Nore/mg/

This is my first map please don't kill me :))


RyanZem:
screenshot
CTF doesn't support farming and bed mods.

I don't think others will like the map because of the forest. Forest maps are boring. I think it will be a bit difficult to capture also.

farooqkz:
@RyanZem those are for decoration. Any way to keep them?
And this map meant to be a village map.

RyanZem:
> @RyanZem those are for decoration. Any way to keep them? And this map meant to be a village map.

CTF doesn't support beds and farming mods.

farooqkz:
okay I'll try removing them

farooqkz:
Hey do I need to rebuild the map for v3?

PR #236, opened 2021-11-14
Adding Sewer Under Village

winniepee:


sewer_village/map.conf
```diff
@@ -0,0 +1,22 @@
...
+base_node = ctf_map:meselamp
+others = This map was made by realyg, liverpool, winniepee0, DiamondPlane, Elvis26, j45, and gameit.
+license = CC BY-SA 4.0
+phys_speed = 1.5
+chests.1.n = 80
+chests.1.from = (-110,-70,-110)
+chests.1.to = (100,-50,100)
+chests.2.n = 80
+chests.2.from = (-110,-25,-110)
+chests.2.to = (110,50,110)
```
Panquesito7, line 22: Missing an endline.
```suggestion
chests.2.to = (110,50,110)

```

MoreKat (commented):
Basic but potentially fun map. Kind of a mix between Snow Village/Plain Battle and Mayan Legend/Water Academy
Bases are horribly unbalanced though. One base above ground and one underground, that can't be fair. I know you did this because you want to prevent that the underground section won't be used, but IMO that could be done by making larger and more entrances to the sewer so players can literally see the tactical advantage of using those to capture. You could also add a treasure in the sewer, or have both bases not completely underground but lower than the surroundings (like in Xenon's swamp/jungle map - which was removed - where both the underground and above ground environment were used)

HobbitPower:
Closing this PR, new one can be found at #281

PR #239, opened 2022-01-10
Adding "Battle of Hoth"

winniepee:


TSafa-23:
Be sure to add screenshots, a short description and license

battle_of_hoth/map.conf
```diff
@@ -0,0 +1,20 @@
...
+author = winniepee0
+time_speed = 1
+enabled = true
+name = Battle of Hoth
+map_version = 2
+size = return {["y"] = 140, ["x"] = 230, ["z"] = 230}
+start_time = 5900
+skybox = none
+hint = 
+treasures =
```
Panquesito7, line 17: I don't think you should add this if it's empty. It could mess up stuff more. 😅

TSafa-23: It probably has a nil check, I wouldn't worry about it

HobbitPower:
First of all: the structures are looking stunning. The map itself reminds a bit of the snowland map, but the base positions are better. The base ground is automatically made out of unbreakable cobblestone, don't know if you are planning to keep it this way or if you want to change it to unbreakable snow etc. 
One question about the mese/diamond mines, for me its too unbalanced: red side got "only" 37 mese & 12 diamonds BUT blue got 64 mese & 16 diamonds.
In v2 me was told to upload a screenshot with a 3:2 ratio. Dunno if this is still needed in v3. If yes you should update your screenshot.
Some chests spawn 1 block in the air, because you used the "thinner snow slabs". Don't know if this is a problem, i am fine with that. 
Otherwise everything worked fine and i found no bug yet.

HobbitPower:
Good so far, but i discovered one major problem: you forgot to add the team chests (you have to place them manually now in v3).

Btw red got 2 more mese than blue now, but its almost balanced now so 👍.
Idk if you want to ask LV (for example) about the screenshot ratio. In v2 3:2 was needed. But i have honestly no idea how this is handled in v3.

HobbitPower (approved):

MoreKat (commented):
The structures look nice, but provide barely anything for the gameplay. Boring terrain, very similar to maps we already have (I'm thinking plains, coast, two hills, snow land). Large portion of it is just flat. I also think you should add a hint about where to find the ores.

PR #240, opened 2022-02-04
Sam's Mansion (Updated for v3)

xenonca:


HobbitPower:
* chest zone is needed.
* the blocks next to the red barrier are breakable in building time (you can stand there after breaking, but ofc you get teleported back). had a similar problem with my map, dont know how to fix it...
* otherwise everything else worked fine + great that both sides got mese now 👍

MoreKat (commented):
Like HP said, needs chest zones. Also I'm not sure if this map is a good fit for the mode Nade Fight, as it's quite small. I think you need more space to be able to dodge grenades, especially when it comes to capturing. Please let me know if you think otherwise!

xenonca:
> Also I'm not sure if this map is a good fit for the mode Nade Fight, as it's quite small. I think you need more space to be able to dodge grenades, especially when it comes to capturing. Please let me know if you think otherwise!

I honestly don't know. I'd say there could be enough space to hide under the furniture and behind walls/secret passages. But well idk, I think the best way to find out is by trying, should be easy to disable it afterwards if it really turns out a mess.

PR #241, opened 2022-02-09
Abandoned Battlefield Update (CTFv3 update: 4 Teams Map)

HobbitPower:
* updated map for CTFv3
* turned the old map into a 4 Teams Map (added second red barrier, made 2 more shelters breakable)
* few changes to the map (new ground in the middle, added 4 bigger flags, partly prevented tunneling)


olliy1or:
I like this update
now you can actually get out of the water even if you don't have blocks lol
Maybe the two teams on each other might target each other before going to the other side though :thinking:. but i like it, looks alright

olliy1or (approved):

LoneWolfHT:
Make sure you provide ores if you want to enable this for `classic`

HobbitPower:
> Make sure you provide ores if you want to enable this for `classic`

i would have to add 2 more mines + reopen the other 2 (i closed the current ones). not the biggest effort but right now i am fine with just enabling the classes + nade mode ^^

MoreKat (approved):
Nice! My only remark is that it might be a bit too dark, with the skybox. It shouldn't be a problem unless players complain about it, then you might have to take action.

PR #242, opened 2022-02-11
Volcano сonversion to version 3.0

ksandronline:
Im check in my local PC.


Panquesito7 (changes requested):
You should maybe update the screenshot of the map if you updated it. 🙂

ksandronline:
> You should maybe update the screenshot of the map if you updated it. slightly_smiling_face

I added plates with the names of ships to the ships. But the screenshot will not display this. There are no significant visible changes to the map.

ksandronline:
There are no team chests on the map, I didn't know that they had to be added manually. in progress...

ksandronline:
Im add team chest`s. Map finish and ready to upload to game.

MoreKat (commented):
Looks good, but I suggest adding to the hint that the chests can be found _in_ the ship itself. Otherwise people will go looking across the map (as they're used to), and not find any during build time.
So something along the lines of "Look for the chests in the airship and a secret treasure inside the volcano!"

ksandronline:
Beginners playing for the first time on the map, probably such a hint would be useful. But there are a lot of glasses on the ship through which you can see these chests inside the ship. I don't think it's necessary to focus on those who are playing for the first time.

I didn't know about the names of the ships, it's a pity. The names were Lucy and Olga.

MoreKat (approved):

PR #245, opened 2022-03-24
Add new map "Dance of the Dragons"

HobbitPower:


MoreKat (commented):
Fun, original and beautiful map! Clearly a lot of work has gone into it. I like all of the secret, hidden rooms. The treasure might be a tad hard to find at first, but after some time players will find it blindly. Speaking about blind, my only concern is that some parts, especially where the obsidian is, seem to be very dark. I can't see much, at least less than I would like to see when I'm escaping with the flag. So I suggest you add some light sources on the outside of the volcano, I think flames would fit very well? This is the only change I propose.

P.S: +10 for the poem in `others`!

MoreKat (approved):

olliy1or:
A lot of lava but a beautiful and original map :+1:

PR #246, opened 2022-03-25
Add new map "Wonderland"

ksandronline:
I have created a new map. I took the Сaverns map as a basis. I replaced all default:stone blocks with default:glass. Then I replaced 1/40 of the default block:glass on wool:white. Added shotgun + 10 ammo to initial_stuff. Slightly reduced the speed and gravity, and increased the height of the jumps. The highlight of map is that shotgun destroys glass and wool very much. Now you can almost not use a pickaxe. example - https://youtu.be/HyzSnN4ik3U


ksandronline:
see #248

PR #248, opened 2022-03-27
New wonderland request

ksandronline:
I have created a new map. I took the Сaverns map as a basis. I replaced all default:stone blocks with default:glass. Then I replaced 1/40 of the default block:glass on wool:white. Added shotgun + 10 ammo to initial_stuff. Slightly reduced the speed and gravity, and increased the height of the jumps. The highlight of map is that shotgun destroys glass and wool very much. Now you can almost not use a pickaxe. example - https://youtu.be/HyzSnN4ik3U


MoreKat:
@ksandronline So I can close this PR: [https://github.com/MT-CTF/maps/pull/246](url) ?

ksandronline:
yes of course. please.
I'll do it now.

HobbitPower:
Wrote it at Discord already, so once again here: 
I am against adding this map, because imho it seems confusing with all the glass. Would loose interest in playing a match there probably pretty soon. For me all the glass is not pleasing to the eye.
A powerful item like the shotgun as a spawn item + the jump effect would also annoy me in the length of time. Got also the feeling that matches there could result into killfarming, without attempts to get the flag. 
Some others also mentioned the possibility of lags due to the amount of glass. 
I prefer the original caverns map, see no need to add this one.

PR #252, opened 2022-07-18
Desertic Coast

LuismiInTheZone:
Here is the new CTF map. Around here there are the files...

Let's see if finally this map can get into the game.


bloopy2 (changes requested):
ff

Desertic Coast/map.conf
```diff
@@ -0,0 +1,20 @@
+game_modes = return {}
```
bloopy2, line 1: make this `game_modes = return {"classic", "classes", "nade_fight"}`, or remove this map from whatever mode you think this won't be good on

PR #262, opened 2022-08-07
Add accessible ores to Two Hills and Two Lakes for Classic mode

MoreKat:
I added 36 iron ores to each side. On one side I created a small cave to hide them in. On the other side I put them in an existing cliff. I also added a hint about where to find these ores.


HobbitPower (approved):
Nice little update. Classic will be more fun there and it fits to the OG map.

PR #264, opened 2022-08-19
Give both teams a similar amount of apples in Coast of Nostalgia

MoreKat:
Before there were **33 apples to find on red's side** and around **172 (!) on blue's side**. I've brought this back to 93 apples on red's side and 108 apples on blue's side, with as little structural change as possible. I've only placed some more trees around the spots where there were already trees on red's side, and made a small island behind red's base, next to the map border, where I've planted more trees. This should not change the gameplay.


magnetar47:
You can't self-approve your PRs, lol. That's cheating. :P

MoreKat:
> You can't self-approve your PRs, lol. That's cheating. :P
>
Why not? I approve this change as a map manager, not as author of the PR. And given that I'm one of the only two active map managers it's almost necessary, since this needs two approvals.

HobbitPower (approved):
Looking good and it's now much more balanced

PR #268, opened 2022-09-26
Update "The Legend of Capturing" by gameit for v3

HobbitPower:
Changelog (by gameit):
- added some coal and iron for classic
- added a skybox
- made blue's roof digable
- fixed a wall
- already tested the map


MoreKat (commented):
Looks good, one question: why did you disable it for Nade Fight?

HobbitPower:
> Looks good, one question: why did you disable it for Nade Fight?

gameit thinks that the height differences between the two team bases are too different and in nade-mode high-ground is too much of an advantage. That would make the map too unbalanced there according to him.

MoreKat:
> > Looks good, one question: why did you disable it for Nade Fight?
> 
> gameit thinks that the height differences between the two team bases are too different and in nade-mode high-ground is too much of an advantage. That would make the map too unbalanced there according to him.

Okay, makes sense

MoreKat (approved):

PR #269, opened 2022-12-14
Add new map 'Obsidian pyramids' by ubuntu22

MoreKat:


LuismiInTheZone:
I tested the map and I don't see any errors, maybe that some places are easier to get in than others... But maybe that was your plan... But in general everything seems fine...

ColorBlob:
I'm not done testing your map yet, but I did find a large portion of the bottom blocks breakable, making it easy for players to dig straight down and escape the map. Otherwise, I appreciate your creative map design :)

ColorBlob:
I did not find any more problems with the map, but I did find a few minor issues. 

- Team red has more ores on their side when compared to team blue (unless I missed something). This gives team red an unfair advantage. 

- When spawned in the map, the player is automatically placed to be facing a bit towards the right instead of facing forward.

On a personal standpoint, I like how the bridge from both team’s base leads to the central pyramid, encouraging battle between the teams, and I like how the pyramids on the side of the map provide cover for flag capturers. But personally, I don’t like the look and feel of the map.

LuismiInTheZone:
> I did not find any more problems with the map, but I did find a few minor issues.
> 
> * Team red has more ores on their side when compared to team blue (unless I missed something). This gives team red an unfair advantage.
> * When spawned in the map, the player is automatically placed to be facing a bit towards the right instead of facing forward.
> 
> On a personal standpoint, I like how the bridge from both team’s base leads to the central pyramid, encouraging battle between the teams, and I like how the pyramids on the side of the map provide cover for flag capturers. But personally, I don’t like the look and feel of the map.

Both teams have same ores, I just looked to confirm

ColorBlob:
> > I did not find any more problems with the map, but I did find a few minor issues.
> > 
> > * Team red has more ores on their side when compared to team blue (unless I missed something). This gives team red an unfair advantage.
> > * When spawned in the map, the player is automatically placed to be facing a bit towards the right instead of facing forward.
> > 
> > On a personal standpoint, I like how the bridge from both team’s base leads to the central pyramid, encouraging battle between the teams, and I like how the pyramids on the side of the map provide cover for flag capturers. But personally, I don’t like the look and feel of the map.
> 
> Both teams have same ores, I just looked to confirm

Sorry, where did you look for the ores? **_The ores in the lava are not the only place where there are ores_**. Ignoring the ores in the middle lava, red has 22 diamonds and 99 iron, while blue has 18 diamonds and 74 iron. This should be the case unless I missed another place containing ores.

LuismiInTheZone:
I don't know if I'm wrong because there are a lot of minerals, but the blue team has 5 mese blocks, while the red team has 4 mese blocks.

Personal opinion: I think there are a lot minerals, I would reduce them a bit

LuismiInTheZone:
Now everything is ok, there are no problems

LuismiInTheZone:
Ok, seems everything is ok, now getting ores is easier and the lava ones have been removed... I only have a suggestion to make: some chests appears at old rooms of diamonds and mese, and it's difficult to get into those, also some chests spawn at the start of the volcano
Minetest 5 6 1  Singleplayer   OpenGL 4 6 0  19_04_2023 20_23_28
Maybe is difficult and you think isn't necessary, it's only my opinion, but I would delete some spawns from chests, but once again, maybe it's difficult and isn't worth it to fix it.

lazylier:
I like this map overall, except that the red barrier is not in the center. Red's side is 10 nodes longer than Blue's. Another thing is, the wooden structures in the corners. What are they supposed to be? This map doesn't have any obvious theme, but it looks like it would be fun to play anyway.

LuismiInTheZone:
Ok, I'm going to say what I think about the map since there are no errors:
My opinion:

For me, it's a cool map as there aren't any maps like it in the game right now. But, for example, things I would change is that to take some ores, it's hard to get to them (especially mese blocks), since there's a lot of obsidian, and it's very hard to destroy, so you could do some kind of tunnel to get to them, because if not, it's going to take a long time. And what HP said some time ago, trees don't fit with the pyramids biome. Well, this is kind of what I think, but you don't need to change these things, especially the trees, as they can be whatever you want.

LuismiInTheZone:
I love the map, you fixed a lot the things and now is very cool. My opinion now is this:

- The base looks a bit dark, did you thought in putting in it some light?
- I found one thing that at least for me disturbs me a bit:
mm
mmm
- Also I found that keeps being some water that maybe you missed, look:
water

Seems cool to me, I'll wait for other reviews, very cool map with those new tunnels

src4026:
I think this needs an update with the latest Map Editor?

Themostrandom:
I found 3 things: 
chests spawn above lava (spawn zones need to be redefined)
Added lava-filled tunnels (with the possibility of plugging them with ladders)
And give access to this obsidian sphere or remove it (no minerals or chests present).
screenshot_20240409_205156
screenshot_20240409_205130
screenshot_20240409_204956

SoulS33ker:
1. ore imbalance
   iron 61:76
   mese 12:16
   diamond 15:17
   two dia ores found in area of red barrier
2.red team seems to have a tail
screenshot_20240729_195327
with these fixes i believe its a good map to play ^^

ClothierEdward:
I like where you're going with this map. It reminds me a lot of the Dance of the Dragons map and the Volcano map but is still distinct from both of those maps which I quite like.
That being said, there are some things that do need fixing up.

First of all, some of the pyramids generated kind of weird. They're either cut in half or cut off by the wood corridors as you can see here.
screenshot_20250202_173759
screenshot_20250202_173751
screenshot_20250202_173512
Second of all, what are these stone platforms for? Are they supposed to be used for sniping in classes or are they for something else?
screenshot_20250202_173628
And third of all, a minor nitpick but I wish the shape for the roof of the team bases was a bit different. Maybe add another dome on it or change it to be shaped like a pyramid? 
screenshot_20250202_173527

That's what I've found though. Fix those issues and it would make for a pretty good map.

PR #272, opened 2022-12-27
Add new map "Savanna Ravine"

a-blob:


LuismiInTheZone:
Everything looks fine, looks good and has a new design of the ravine...

Just one thing: Don't you think chests could be around the map also?

LuismiInTheZone:
Everything seems fine and there's no problems, ores are ok...

Just some little thing, don't you think chest could also be added around the map?

farooqkz:
Looks good but I've got an idea which might improve the gameplay: Having bases in the opposite corners or at least closer to the corners.

ColorBlob:
I will update the chest zones when I have time. As for the team base, I will keep them like that and not make them further from one another **_for now_**, because It's already pretty hard to capture and increasing the distance will not help.

a-blob:
Changes: made ore/chest room easier to find, made it impossible to jump across the ravine without building.

HobbitPower:
The map is well done and everything seems to work at first glance. Good decision to make the treasure room at least a little more obvious.
I'll remain neutral for now, as I want to wait for more reviews. But on the whole, I'm positive. Will review it again soon. Good work.

src4026:
The map is quite good. It gives me the open feeling of the Plain battle combined with the wall. The ores are distributed in both the treasure area and the cavern, which might make the ore mining a bit more fair for the players. Overall, the map is designed well. I didn't find any obvious issues. Nice.

MoreKat (commented):
Fun map. However, I feel like there are not enough chests above ground. I would add some more, especially because I think the treasure room is still hard to find. Why don't you add a ladder to it, covered by a block? The players would only have to find and dig the block then. I remember it was like this in the map Lava Puddles (not that you should take that one as an example). Other than that, I like it and I look forward to seeing it on the server.

src4026:
> The map is great. It gives me the open feeling of the Plain battle combined with the wall. The ores are distributed in both the treasure area and the cavern, which might make the ore mining a bit more fair for the players. Overall, the map is designed well. I didn't find any obvious issues. Nice.

I haven't checked it by making a local server and playing it. Might have to look into that for the game mechanics.

a-blob:
I made the chest/ore rooms more accessible by adding a ladder.
@MoreKat do you think I still need to increase the amount of ores above ground?

MoreKat:
I think it's fine.

MoreKat (approved):
It's a simple but fun map. I approve.

HobbitPower (approved):

PR #273, opened 2023-01-01
New Map: Greek Temples

winniepee:


LuismiInTheZone:
@winniepee I'll test up the map but... Why you keep adding new maps and don't edit to v3 the older maps?

LuismiInTheZone:
The map is ok, a bit small maybe... But seems everything is fine

winniepee:
> @winniepee I'll test up the map but... Why you keep adding new maps and don't edit to v3 the older maps?

I'm busy with school, didn't have that much time on converting maps that I made. (Might do one on March)

ColorBlob:
Hi, I tested your map, and I really love the map design and the idea of putting teams close to each other on flat terrain. Seems like a fun map.
I did notice some minor/non-serious problems.
- There is a strange hole behind orange's base in the corner. Did you forget to fill it up?
image
- One of the team's chest zones has 2 more chests compared to the others. I don't know if it's intentional or not.
- The screenshot's ratio is close to 3:2, but **not** 3:2 (unless my calculator is broken).

Personally, I don't think it's a good idea to have red team together with orange team, since it's pretty easy to confuse both teams due to their similar color.
__________________________________________________________________
Here is my personal opinion: 

This map is a bit small, and together with the flat terrain's help, it's totally possible for matches to end within minutes. 
Because of the flat terrain and short distance between the teams, It's going to be chaotic as teams try to capture/kill their enemies; since the flat terrain and short distance make this map a paradise for block spammers, spawnkillers, shotgun spammers, and speedrunners. 

Again, all of that ^^ is my personal opinion.

LuismiInTheZone:
@winniepee why did you delete the repository?

winniepee:
> why did you delete the repository?
Had some issues with the PR ...

MoreKat (commented):
Interesting map. It's very small, so the matches will probably be rather short and intense, but I don't think that's a problem. It adds some variety. I agree with Greenblob on the colours, I suggest switching orange or red for green. Orange and red are too similar and will cause confusion. I also suggest adding blocks to build a base with to the initial_stuff, maybe even reinforced stone or damage cobble, because it will be vital to have a good base with there being such a small distance between the teams. I would also add a hint about where to find the chests (in the ceiling), so the players don't start looking elsewhere and waste valuable time that they need to prepare. Lastly, I wonder why you didn't enable this map for the Classes gamemode? I think it would be fun in that mode as well.

winniepee:
> I wonder why you didn't enable this map for the Classes gamemode? I think it would be fun in that mode as well.

Will consider enabling Classes. Still gonna disable Nade Fight though, I'm scared that the match will end quickly.

MoreKat:
> > I wonder why you didn't enable this map for the Classes gamemode? I think it would be fun in that mode as well.
> 
> Will consider enabling Classes. I still won't enable Nade Fight though, I'm concern that the match will end quickly.
> 
> EDIT: Just realized that my greek_temples branch is now gone for some reason ... There's no way to upload the updated files.

Make a new PR then?

winniepee:
Nope. Won't work.

winniepee:
New PR #284

HobbitPower:
> New PR #284

Closure of the PR due to conflicts. New can be found there ^

PR #278, opened 2023-02-23
"The Olympics" CTFv3 Convert

winniepee:


LuismiInTheZone:
Well the map is amazing and I played this before as you can find the map in v2 (and here on github)... And I saw that you changed the bases of their places, I guess it's because of Kat's comment. .. Well, I don't know what to say about what she said a long time ago ...

But this is what I saw:

- The points of the chests are not well placed, because look at this image, I recommend looking for the positions again and doing it

Minetest 5 6 0  Singleplayer   OpenGL 4 6 0  23_02_2023 21_24_23

- Some things are missing from the .conf file:

HobbitPower:
> _ No description provided. _

Just for the record: This PR is a revised version of PR #224. Some already existing feedback can be found there.

HobbitPower:
It is a nice idea, but I don't like the implementation. 
The biggest problems for me are still the bases and their positions. They are too unbalanced, differ too much in their protective function. Also in terms of height and distance to the centre. 
Otherwise, the terrain is rather flat and the buildings seem randomly placed. 
I will therefore not approve the map.

And for the future: Please do not select the entire map for the barrier area. This only applies to 4-team maps (maps with several barriers). Otherwise: use the smallest possible area that contains all red barrier nodes.

PR #279, opened 2023-03-02
Add new map "The Spring Airport" by MrMars

HobbitPower:


ColorBlob:
Presently I did not find any problems or bugs with the map.
This map is similar to Abandoned Battlefield. In other words, it's just Abandoned Battlefield cut in half and all the tanks are replaced with planes. 
The overall experience while playing the map is similar to Abandoned Battlefield, except that now the other two teams are gone and there's Classic mode. Also there's a secrete tunnel near each team's base that seems to encourage players to dig all the way towards the other team's base, which makes this map a bit unique.
To quote -sniper-, 
> "Map should be unique and not similar to other maps... the design of a map should encourage differing game play and tactics."

From what I see, this map mostly encourages the same gameplay and tactics as Abandoned Battlefield, making it boring.
I like the design of the tower though :)

MoreKat:
I agree with a-blob. This map is not original. If you don't want to make a map from scratch, I recommend taking one of the maps that were disabled after a map voting and improving it (maybe you can ask the community for advice). Ksandr also mentioned that they needed help replacing the buildings in their maps that they used schematics for (which were removed because of license issues).

HobbitPower:
As mentioned above, the map has similarities with already existing maps.

PR #281, opened 2023-03-10
"Sewer Under Village" CTFv3 Convert

winniepee:
Just changed few things before converting this.


a-blob:
Seems like a fun map. I did not find any issues. Why is the map author Map-Makers Inc.?
I don't think it's a very good idea to make the base totally unbreakable except for a two block hole. **But if you want it that way then it's fine.** I can just imagine players blocking the hole with 10 doors and tons of wooden planks. It's going to be a bit hard for players to go into their enemy's base. Before they even successfully break the doors and blocks blocking the two block hole (which takes a long time to break), they are going to get killed. Once they are killed, the doors will be placed back by the base campers. Rounds are likely going to last longer than _The Wall._

winniepee:
>Why is the map author Map-Makers Inc.?

It was when we had a group that specifically make maps. We just joke around and named our group "Map Makers Inc." (Might change a bit)
>I don't think it's a very good idea to make the base totally unbreakable except for a two block hole.

Yeah ... Do noticed that. Might make the glass breakable rather than unbreakable.

HobbitPower:
Since you're the author you could have also updated PR #236

HobbitPower:
Not my final review: Very beautifully built map. Good that both bases are now above. I also like the double wall in the middle, it allows interesting tactics, since unlike The Wall you can also go around and you are able to use the underground. 

I would almost go one step further than GreenBlob and make the whole base out of breakable blocks. But that's up to you.  

I could imagine that the sewers are used less, because they are very far down and it takes a while to reach the bottom. The chests and ores are at least a reason to go down. I don't know if you want to move the sewers up a bit, a lot can go wrong with WorldEdit. I think it's fine anyway, fighting is more above ground anyway. Maybe the entrances to the underground are still not obvious enough, but at least there are several.

I will test it again in the near future, but I for one see a good chance for the map.

winniepee:
I'm just gotta keep it as usual. I think it's fine.

HobbitPower (commented):
After another review: I still think that the house should consist of destructible blocks, at least the roof.

Did you set the whole map for the barrier area? It should cover the smallest possible area, so really only everywhere where red barrier blocks were used.

Unwalkable Dirt is actually not intended for the classic mode. Maybe you'd better replace it with another type of dirt, although I can see the intention behind it.

MoreKat (commented):
First of all, please make sure your screenshot has a ratio of 3:2. The map and its buildings are quite basic, bordering on ugly, but I think it provides the necessary terrain for different strategies so that's a good thing. The map reminds me a lot of the Village map that got removed (I think?). I agree with HobbitPower that the house should be completely unbreakable. I also think that you should specify in your hint what those 'treasures' are (chests). Treasures can also mean ores and I think it's necessary to make clear where you can find the chests because there are no chests anywhere else. Maybe you should change that though, I would add another chest zone on the surface with fewer chests because I cannot imagine that it'd be fun to have 15 players (or more) racing down to the sewer and fight over loot. 
Lastly I'm afraid the phys_speed is too fast. We started off with this speed in another map and the matches ended in minutes. This will especially be the case in Classes, as the medics already have elevated speed. When sprinting they're literally flying.
Once those issues are fixes the map is good to go in my opinion.

winniepee:
Just updated both the map and screenshot. Both houses are now destructible.

omikekimo:
I really like this map, it is well balanced, it has symmetry but not too much, it has a good variety of terrain that will offer a variety of choices in combat and evasion and the bases are pre built so I can just run and gun without worrying too much about speedrunners.. It is I am sure going to be great fun to play and I only have two points of criticism: 1) the stairways down to the sewer need to be made of actual stairs, they are annoying to climb up and down, 2) the name. To me it looks more like an army camp with a bunker underground and a training facility above ground than a village. I might be tempted to put a few more ores but then there are loads of chests so that is a minor point.  What I would definitely also say is that the stairways should have a solid landing at the bottom and not lead into a puddle, in reality you would always see an adequate landing at the bottom of any stair. It kind of just breaks my immersion a bit when the rest of it looks soooo good and just that little bit lets it down. I am glad that this is the first time I have looked at this, it seems very polished and I am looking forward to playing it.

HobbitPower:
> I would add another chest zone on the surface with fewer chests because I cannot imagine that it'd be fun to have 15 players (or more) racing down to the sewer and fight over loot.
 
I agree with that. It would be nice if you can add a chest zone for each team overground. Maybe 10 - 15 for each? That way a visit to the underground is still attractive. 
You can keep the amount of chests there or adjust it. Your choice. 
Maybe it would be good to edit the hint then `(more chests`).

> 1) the stairways down to the sewer need to be made of actual stairs, they are annoying to climb up and down, 

I agree that it can be annoying, although it's not the biggest issue for me. So think about if you want to change it. 
> 2) the name. To me it looks more like an army camp with a bunker underground and a training facility above ground than a village. 

I see the point, but then you might have to change the map folders name as well and I don't know if it's really worth the effort. It can be seen as a training village for urban warfare. So it's still a village ;)
___
So @winniepee adding 2 chest zones would be good and then I would be ready to approve the map. But please also take note of Mike's feedback. Maybe you want to implement something.

After that, all that's missing is a review from another map manager.

HobbitPower (approved):

MoreKat (approved):
Decent map, I think its best feature is the variety of the landscape. It's very ugly but that's not a valid reason to disapprove. I'm not sure if it's a good fit for the mode Nade fight but we will find out when it's played.

PR #282, opened 2023-03-16
Add new map Silver Lands

LuismiInTheZone:
I created this map like 3 months ago, but the Lost mountain stuff was going on for the moment so I waited, when I was going to put the map I thought it wasn't good, so I edited it a bit and built some bases...


a-blob:
That ^ is a lot of commits. 
Presently I did not find any issues with the map. Everything is fine. Ores are satisfyingly balanced. However, I find this map similar to Dessert Spikes. Both maps have the "spikes", and the water below the "spikes". What makes Silver Land different is the bridge you built across the ~~mountains~~.

LuismiInTheZone:
> That ^ is a lot of commits. Presently I did not find any issues with the map. Everything is fine. Ores are satisfyingly balanced. However, I find this map similar to Dessert Spikes. Both maps have the "spikes", and the water below the "spikes". What makes Silver Land different is the bridge you built across the ~mountains~.

Sorry, the commits are from the last lost mountains PR, I don't know how to delete them. Also, this is this 1000 times away from Desert Spikes, that map has taller mountains, and this aren't mountains... Well, maybe yes, but you don't need to climb them

HobbitPower:
Not my final review: Well yes it is a mountain/hill map, so I also see similarities to existing maps (no discussion about the definition of mountains again, please). However the bases are better placed than in Desert Spikes imo.
I assume your intention is that players use the bridges rather than the terrain. Not sure if this will be the case though.

At first glance, the map doesn't really stand out, it's too similar to other maps. The concept with the bridges is a good start, maybe you can build on that and try it in the form of a new skymap? Abandoned Isles also had bridges, but they were too far from the ground, so that a fall almost always led to death. Maybe bridges and platforms on different levels? Just a suggestion, maybe you can find an even better way to edit Silver Lands.
In its current form, I personally just don't find it unique enough.

Anyway I will give it another try soon. 

Also next time you might want to use a new branch for the PR or fork the CTF maps repo again, lots of commits indeed.

lazylier:
I think it's strange that there are only three diamonds, and two of them are right together so the third is nearly useless.

LuismiInTheZone:
O bruh I messed up

LuismiInTheZone:
I'll open this again when I finish the whole update of the PR

PR #284, opened 2023-04-25
Add Greek temples (New)

winniepee:


winniepee:
Ignore the first 3 commits. Just gonna let you guys know ....

src4026:
I feel the map is plain. Don't get me wrong, the temple architecture is top-notch, but the rest of the landscape seems to miss something. I think there is more to this map that can be done. :D

winniepee:
I might consider "expanding" the map, but who knows ¯\_(ツ)_/¯

greek_temples/map.conf
```diff
@@ -0,0 +1,21 @@
...
+phys_jump = 1
+phys_speed = 1
+size = return {y=57,x=229,z=230}
+map_version = 2
+enabled = true
+name = Greek Temples
+phys_gravity = 1
+initial_stuff = return {"default:pick_steel","default:torch 10","default:cobble 50"}
+author = winniepee0
+hint = Get treasures inside the temples roof!
```
MoreKat, line 14: `temples` should be `temple's` (more correct would be `roof of the temple`)

MoreKat (commented):
First of all please use a screenshot scaled 3:2. I really like the map and most of the builds, especially the warrior statues. However, the wooden bridges look a bit out of place in between the temples made out of stone. You might want to change that for a more cohesive style. Also, outside of the temples the map is quite flat and bare. Of course this can be a choice. I don't think this is a reason to disapprove of the map, but for more aesthetic pleasure you could add some kind of rocky natural landscape.

PR #289, opened 2023-05-14
Add new map "Cargo Port" by Themostrandom01

HobbitPower:


LuismiInTheZone:
Everything looks all right, but the only problem (for me) is that the map is too small and easy to go to the other base.
I also keep thinking that where's the base put... that looks like the map abandoned battlefield... But in general I like a lot how you made the buildings, looks very cool but... I don't know, I feel something is missing... 🤔

Now to the .conf file part:

Why not adding classic to your map? I don't know if you have some variation or problem that can affect to classic...  

Keep in there, great job

lazylier:
It's nicely built, but maybe a little small. It's confusing that the message saying to look in the sewers, but there doesn't seem to be anything there to find.

HobbitPower:
Will add the WIP Label, since Terraformer was about to change/add some things. 
(@a-blob I understand what you mean about the wool, but Minetest unfortunately only offers a small selection of red/blue blocks)

src4026:
So I had a look at the map. Everything seems well and good for now :+1:. I found some inconsistencies in the block types, that is, indestructible and destructible variants of blocks as in the screenshots attached herewith.
Screenshot from 2023-06-17 16-18-14
Screenshot from 2023-06-17 16-19-24
/MT-CTF/maps/assets/129506166/20857eb1-9b56-484a-80cd-3dc81c8f7e0e)

Overall, the map is good, and well-made.  ~~The sewer system was quite good in design aspects, mine was just a lazy one, I suppose I should not put too much burden on myself as my map had other things more important~~.

src4026:
I have edited the above comment as there were some issues in the images :point_up:

LuismiInTheZone:
Ok the map looks great, maybe some indestructible bugs, I don't know if you did it like that on purpose... (other indestructible bugs are leaves)
errors
In the image, at the point where there are leaves, from there, I can't destroy them anymore.

Also the hint doesn't make sense, I would recommend you puting something about the ores.

LuismiInTheZone:
Another thing, ores are wrong, there aren't the same ammount in each team part:

Team red: 4 diamonds, 10 mese (indestructible)
Team blue: 11 diamonds, 6 mese (indestructible)

Then you also have the ones of the middle, but those can be taken by both teams

LuismiInTheZone:
You changed almost nothing of what we told you, you only changed the trucks, which are now indestructible, and the minerals, but they are still unbalanced:

red: mese 21, diamonds 10
blue: mese 19, diamonds 14

Please read all of our comments again and correct them, and later, keep the minerals balanced, delete them and put them again correctly if you want.

LuismiInTheZone:
Ok here is the review: the ores are now correct, the problem is that you didn't fix what we told you of destructible/indestructible blocks. Isn't necessary but would be better if you do it as we told you to do everything destructible/indestructible. You can find all our comments again here. Another thing, this is for my opinion: if I was you, I would put a steel pickaxe, because if not maybe is a bit difficult to get all those (my opinion, if someone more can say what he thinks of this wouldn't be bad too). But nice, keep there.

LuismiInTheZone:
And... what did you change?

LuismiInTheZone:
There are still some parts from the truck that are destructible, but for the rest, seems is ok

src4026:
> opinion, maybe isn't necessary since there are a lot of chests. If someone more can say what he thinks of this wouldn't be bad too). 
I think the current stone pickaxe is fine. Also, I'll review the map sometime later.

ScaleRaker:
I definitely liked this map when I tried it out. The only things that I thought was weird was the fact that some parts of the trucks were indestructible while other parts were destructible. Also, the inside of the boat is kind of plain. Additionally, the map could be four-team. I don't know what HobbitPower would say about this, since that would make the map even closer to Abandoned battlefield, but yeah, I feel like that might be good. So, yeah, the map is definitely very nice! 
P.S. The hint in README should not have a space between the exclamation mark and "boat".

HobbitPower:
>Additionally, the map could be four-team. I don't know what HobbitPower would say about this, since that would make the map even closer to Abandoned battlefield, but yeah, I feel like that might be good. 

The way the map is set up, I would rather stick with a 2-team map.
As for the similarities to ABF: there are some, but I would say that the water in the middle, for example, makes a difference. I'm currently neutral towards the map, but I'll test it more closely in the near future.

ScaleRaker:
Hmmm, ok.

Themostrandom01:
When will the map be in game?

Themostrandom:
I have an acount now

MoreKat (commented):
1. Friendly suggestion to capitalise 'port' too in the title, so 'Cargo Port'
2. Pick a better screenshot. Your screenshot should give a clearer overview of the map. Take it from one of the corners with jungle so both base buildings are visible.
3. I think it will be too hard to capture with the building in which the flag is situated being completely unbreakable. I suggest either making it breakable or adding some more entrances
4. I suggest adding more torches to initial_stuff. The skybox makes the map quite dark.
5. I think another good addition to initial_stuff would be ladders. 
6. Maybe consider making more things breakable, for example the structures in wool (not the cranes). In my experience players prefer a breakable map over a map in which everything is unbreakable.
Overall I like the design of the map. The cranes and trucks look really good. The map looks aesthetically cohesive. There seem to be no bugs.

Themostrandom:
I made all the requests, I let you test it, I added sewers under the bases for an additional access, doors and I made the containers breakable.
[Uploading cargo_port.zip…]()

Themostrandom:
[Uploading cargo_port.zip…]()

HobbitPower:
Moving to #334

PR #290, opened 2023-05-14
Add new map ''Beach wars'' by zeroshiki

LuismiInTheZone:


src4026:
The map is set to `enable=false`, I think it should be set to `true`.

PR #291, opened 2023-05-17
add ant war map

lazylier:
This map uses a lot of indestructible terrain to limit tunneling. It is possible to tunnel through the iron areas on the sides.


LuismiInTheZone:
Very good map, it's a good thing to see what you did that players can't cross to the other bases in 10 seconds and make them go to the other part of the map to climb the stairs to the world.

I can't say much about it, just that it looks like the map swamplands but with wool, and maybe, at night, some users won't see much around the surface, so it could bother some players. I like what you did of taking out nade fight since it's a closed ground. And for errors you have a few but in the .conf file, see:

LuismiInTheZone:
Ok, everything is fine, I like a lot the map 👍

a-blob:
Interesting map. I did find a few issues:
- If you touch the red barriers during build time you get teleported back to base. This means that the team zones are off by one block. IIRC team zones have to include the barriers in order to work properly.
- The blocks under the flags need to be unbreakable, but there are two breakable mese lamps. Not sure if this will actually cause any problems though.
I don't think it's a good idea to give the player a stone pickaxe. The stone pickaxe can't dig the diamond and mese ores, but the iron ores are far away from the base. So if the player wants to make an iron pickaxe, they have to dig all the way down to the iron ores. Other blocks like gravel and obsidian might impede the players from digging to the iron ores. There's nothing wrong with this technically but I feel like that it's too hard for the players to obtain the ores, especially considering the fact that there are no hints on where the iron ores are.

I like the idea of the map and the wool flowers builds, but without tunneling, I don't see anything extremely unique about this map. It is possible to tunnel through the iron ores area, but it's hard, especially since that there are blocks such as gravel and obsidian in the way, and that the player only gets a stone pickaxe. (And with tunneling, it's going to be too easy to capture).

a-blob:
I do notice that your conf file says: `initial_stuff = return {default:shovel_stone, default:torch 10, default:cobble 30}`
However I'm not getting any of that initial stuff, so something is probably wrong with the conf file.
image

lazylier:
Thank you for finding those problems. I've fixed them all, except the stone pickaxe is still there because I was thinking it would slow down tunneling if players had to get a better pick from the chests first. The idea is that tunneling should be possible but not too easy. As it is, I'm questioning whether it's too easy to tunnel still.

a-blob:
> Thank you for finding those problems. I've fixed them all, except the stone pickaxe is still there because I was thinking it would slow down tunneling if players had to get a better pick from the chests first. The idea is that tunneling should be possible but not too easy. As it is, I'm questioning whether it's too easy to tunnel still.

It's just my opinion, and as for now I think that maybe it's a bit too hard to tunnel, which will force the majority of players to use above ground. Just a suggestion, but maybe make the grass blocks on the ground breakable so that it's easier to reach the enemy bases?

src4026:
There being only one exit and entrance may be a bit annoying in case the other team's player blocks the tunnels with a hard material. Would it be good to have more paths out and in of the bases? The idea that the map is based off is amusing, indeed :+1:

MoreKat (commented):
I really like the map, it's very unique. Please use a screenshot that gives a better overview of the map. 
The bases are very close to each other so it's hard to find a good balance between 'too easy to capture' and 'too hard to capture'. I don't think that the grass blocks need to be breakable but I do think that there need to be more entrances/exits. Just an idea that I had is to add a maze in between the bases (in the center of the map). You can also connect the exits of each team (where the ladders are) with tunnels. (Specifically that part because the tunnels will be the longest there). If you're worried about tunnels making it too easy to capture, you can add things to slow players down like a layer of water or partially block them. I understand that you want the surface to be used so maybe add more entrances that can only be accessed via the surface? 
It's good to keep in mind that we can always adapt the map after adding it in-game, when we experience that the matches end too fast or last too long.

lazylier:
I've added some tunnels to guide players to the breakable parts of the walls. They're intentionally left dark.

a-blob:
> I've added some tunnels to guide players to the breakable parts of the walls. They're intentionally left dark.

From Discord: Maybe you should also add something like a maze as Kat suggested? (I was a bit disappointed at first when I first saw the PR because an "ant" map doesn't even have many ant tunnels.)

src4026:
@lazylier It would be a good to merge the commits from the main map repository to your fork so that there aren't conflicts and it's up-to-date :+1:. Currently checking our your map.

src4026:
I think you have done good with the tunnels. However I feel that since it being an ant map there should be more, but I would like to know how other's are feeling about it too. Also, should the chest zones be increased to accomodate more? I feel there are quite less at the moment considering it's a four team map. Keep up the fantastic work :ok_hand:

MoreKat (commented):
Good job on the tunnels! I think it would be even cooler if the smaller tunnels connected in the center of the map. You could have an extra room down there where all of the tunnels meet that can be decorated as a storage room (like in a real ant nest, where the ants store their eggs or their food). I understand that you don't want to add too many tunnels so it doesn't become too easy to capture and so the surface will still be used. I think at least one more entrance to the bases is still needed though. Currently there is only one pre-built, easily defendable and blockable. I think it will be hard to capture. As I've said before, you can still change it afterwards if you find it to be too easy to capture with an extra entrance added.

HobbitPower (commented):
This is a nicely built map. I agree with the previous speakers that more smaller tunnels could be useful. If you put the nest suggested by Kat in the middle, the tunnels could lead to the iron ore deposits, for example, so that the tunnels are only exposed once all the iron has been mined. But they don't have to run in a straight line, I'll leave that to you alone, whether and how you want to implement it. The nest could also have extra chests, in which case you would have to add another chest zone. But that's up to you.

Speaking of chests: I think 20 are enough per team, but 25 wouldn't hurt either. But that's up to you again, we can always adjust that quite easily.

And yes, one (or even two) more entrances to the bases would be useful and necessary. They could also be smaller and more hidden. Maybe one can lead down and one up. Maybe with a few obstacles, detours or partially flooded by the rain. I'm sure there are some creative solutions.

Otherwise, I'm looking forward to the map as soon as the problem with easier access has been solved.

lazylier:
> I think at least one more entrance to the bases is still needed though. Currently there is only one pre-built, easily defendable and blockable.

> And yes, one (or even two) more entrances to the bases would be useful and necessary.

Do you mean another tunnel leading to the flag room, or another way through the unbreakable walls, or both?

MoreKat:
> > I think at least one more entrance to the bases is still needed though. Currently there is only one pre-built, easily defendable and blockable.
> 
> > And yes, one (or even two) more entrances to the bases would be useful and necessary.
> 
> Do you mean another tunnel leading to the flag room, or another way through the unbreakable walls, or both?

Another tunnel leading to the flag room that doesn't start at the ladders.

lazylier:
Ok, I've tried adding some more rooms. Each team has another small chest zone with 4 chests, and 12 mushrooms. I'm not perfectly happy with the positioning of the rooms, but I haven't thought of a better alternative.

src4026:
I think the rooms are good at present. But I suppose making them more hidden can make the gameplay different since there would be the visible ones on the top of the map and then the hidden ones underground that can maybe be found by digging a particular block and then some sort of hint. Is the positioning of the rooms in terms of their accessibility/visibility that you are unhappy with? Also, a suggestion based on the ones mentioned previously by others in this thread, what if you combine a type of maze along with the chest rooms and connection to the outer pathway? And, another pathway is needed as mentioned by the others. Thank you for taking my feedback. I am confident that you're thinking of other ideas too for the map tunnels and ways to make it safer with respect to equal advantages for the teams in capturing/preventing a capture.

HobbitPower (commented):
Although no other entrance has been created directly to the base, you can now dig your way from the main entrance to the 2 new side rooms. This opens up more possibilities for conquering the flag.
Of course, more paths and rooms could be added, but from my point of view, it's okay. 
I will test it again soon, but overall I see a good chance for the map to be added.

MoreKat:
I find it hard to judge whether the capture difficulty is at an okay level in this map. I do think it's playable though so I suggest adding it to the game and awaiting feedback. I assume you're willing to adapt your map if the matches last too long (too hard to capture) or end too fast (too easy to capture).

src4026:
> I find it hard to judge whether the capture difficulty is at an okay level in this map. I do think it's playable though so I suggest adding it to the game and awaiting feedback. I assume you're willing to adapt your map if the matches last too long (too hard to capture) or end too fast (too easy to capture).

Indeed. We all alone cannot be able to judge the mechanics of it from singleplayer.

HobbitPower:
I will approve it, but @lazylier be prepared in case the flag is too hard or too easy to capture ;) Then maybe a few changes are needed.

lazylier:
Thank you! I definitely will make changes if there are problems with it. Seeing how it plays should help with that.

src4026:
I feel it is a bit hard to tell which opening is for which team. Maybe some kind of indicator of which team it belongs to - like a flag structure - using which one can tell them apart?

PR #293, opened 2023-05-29
Add Alien Invasion map

ScaleRaker:
This map is a city based map with a few features:
1. UFOs in the sky
2. Subways underground
3. Unique buildings made only for this map


ScaleRaker:
Ill try to work on that.

src4026:
I took a look at the map. First thing, it has a **lot** of stuff to check out!. I couldn't find the exact location of the ores but when I checked there were 10 diamonds on both sides and 0 mese. I wonder why you chose not to keep mese as a treasure ore in your map, or maybe it may be that I missed it. Here is the thing that I noticed are not correct as far as I know:

- The map should be an indestructible cube that should be *inescapable*. I would recommend you to replace the sides and bottom of the map (that are not the indestructible glass) to indestructible blocks.   
Screenshot from 2023-06-03 20-14-24

I have faith that the architecture that is in the map doesn't have major issues. But this may need futher testing. To be honest, one can make a map about every detail of this map :sweat_smile:. 

One more thing I would like to add to this first review is about the bases you've made for each team. They contain glass panes, glass doors, etc. These are easily breakable by shotguns and could be a problem. Replacing them with their obsidian varients obviates this issue. The only change is that obsidian glass won't be breakable by hand, but an implement such as a stone pick (or any tool of a higher level). 

PS: Probably add some flowers to the grass :P.

src4026:
As for the border and bottom thing, I am not sure about how it would affect gameplay in the server as in singleplayer we have to play maps in `singlenode`. I assume it is the same case with the server and therefore, the map should be inescapable. Perhaps someone more knowledgeable than me about maps can answer that :).

ScaleRaker:
Yeah, I didn't see that mistake. Ill fix that. I will also put flowers in the map. I was hoping to find some sort of decor to add the lawns, but I didn't think of anything. But flowers are good. I will also make it unescapable. Thanks for the review

ScaleRaker:
Ok, I filled up the hole and added the dark moon sky box. Still working on the decor like the flowers, etc.

ScaleRaker:
I will update PR soon.

alien_invasion/map.conf
```diff
@@ -0,0 +1,21 @@
...
+phys_jump = 1
+phys_speed = 1
+size = return {y=132,x=229,z=217}
+map_version = 2
+enabled = true
+name = Alien Invasion!
+phys_gravity = 1
+initial_stuff = return {}
+author = ScaleRaker, extremyst, MrRar
+hint = Ore is hidden in the base!
```
src4026, line 14:
```suggestion
hint = Ores are hidden in the base!
```

src4026:
Tested the map, LGTM. I would like to know from those who knew better whether the name "Alien Invasion!" would be approved or "Alien Invasion". Although IMO it doesn't matter much, it could depending on how the standards of map naming are set. ~~I may also be asking this to add to the Map Maker README perhaps. ^^~~ . I like the skybox you've added. Overall, nice! :+1:

MoreKat:
I agree that the name would be better without exclamation mark. None of the other map names contain punctuation

ScaleRaker:
Ok, I updated both of those. There is no exclamation mark now. I also fixed the grammar mistake.

ScaleRaker:
Hobbit Power, I did try to fix it in case you didn't know. If you already knew, then pls ignore this, xD

HobbitPower:
Temporarily WIP, because ScaleRaker wants to rework the map (due to low FPS on the JMA server).

ScaleRaker:
I really did want to change the map to make it better FPS-wise, but I really don't know any way to. Also, the lack of evidence makes me unsure if I should even do anything about it.

ScaleRaker:
Honestly I'm a little embarrassed that there were that many mistakes. Hopefully I fixed all of them now..

HobbitPower:
> I really did want to change the map to make it better FPS-wise, but I really don't know any way to. Also, the lack of evidence makes me unsure if I should even do anything about it.

We could look at how the map performs on this server. If it is lagging too much or the FPS rate is too low, we may have to remove the map again.
There is always a risk with maps with many nodes, especially city maps. But I think your map will do better than the Karsthafen map, for example.

ScaleRaker:
All right. Sounds good.

HobbitPower:
@ScaleRaker 
* Please adjust the screenshot. It should show as much of the map as possible. You can also use one of the following. 
screenshot
screenshot

* I would actually prefer to have a chest zone for each team zone so that both teams have the same number of chests available. 

Otherwise, adding the map is fine in my opinion. A city map was requested by some. 
However, as already mentioned, we will have to see if there are too many lags on the map. You could certainly remove a few UFOs or shorten the underground to reduce the number of nodes, but I don't know yet how much that would really help in the end. The map would therefore be given a "trial period".

ScaleRaker:
Sorry about that. I never can seem to get the screenshot right..

HobbitPower (approved):
Okay. For my part, I approve the map as long as you agree to make any changes if necessary. As I said before, we are testing the map first to see how it can be played in terms of lag etc. 
Be ready if another map manager has suggestions for changes, for me the map is alright so far.

MoreKat (approved):
Really cool map! I like the level of detail. I agree that we should at least try it and see how laggy it gets. If necessary you can perhaps remove a big part off the bottom of the map, it doesn't really add anything. I think the map will also be fun in nade fight, you can bounce from building to building :D

PR #294, opened 2023-05-31
Update map.conf of Lost Mountains

LuismiInTheZone:
Well, after a votation, I have decided to put Lost Mountains for nade fight, even my self hating the mode, I think people prefer it. 
(https://discord.com/channels/447819017391046687/448136308108296213/1113150803197304842) Votation won 12 vs 1 so... lets start the change.

### DIFF
--- a/lost_mountains/map.conf
+++ b/lost_mountains/map.conf
@@ -1,4 +1,4 @@
-game_modes = return {"classes","classic"}
+game_modes = return {"classes","classic","nade fight"}
 barrier_area = return {pos1={y=0,x=117,z=0},pos2={y=152,x=117,z=266}}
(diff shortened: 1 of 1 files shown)


lost_mountains/map.conf
```diff
@@ -1,4 +1,4 @@
-game_modes = return {"classes","classic"}
+game_modes = return {"classes","classic","nade fight"}
```
MoreKat, line 1: Pretty sure it should be `nade_fight` (you forgot the underscore)

LuismiInTheZone:
I don’t like nade fight, this was a bad idea, I don’t want a chaos of suicides

PR #295, opened 2023-06-01
Update map.conf of Big Ocean (Revision of the chest zones)

HobbitPower:
Chest zones increased from 2 to 4 for better allocation of chests. Number per team slightly increased (from 22 to 28 per team), can still be adjusted.  Kept the older format of chest zones, as there were problems with this in the past and chests could not be opened. Please check if this works when testing.


src4026:
Everything seems perfect. No issues. :+1:

MoreKat (approved):

PR #296, opened 2023-07-01
New map: Terraria 

sonicSMASH:
It's by far the largest map in the repository. It's on here for testing purposes; some things like floating trees will be fixed later.


LuismiInTheZone:
Woa, did you make another branch in your personal repo @sonicSMASH ?

sonicSMASH:
It’s fixed now

HobbitPower:
> It’s fixed now

just the right name for the screenshot is missing. delete the current one, rename the file before uploading and upload it again.

HobbitPower:
@sonicSMASH Once again here:
Please check your teamzones again. I was able to cross the red barrier, but got also teleported back to my base while just walking near the left border of reds territory. So please check **all** the team zones. This has happened before with the Race Day Map. So maybe you could test your new map yourself before uploading it.

Also there were chest spawning above the ceiling. So chest zones are messed up as well.

And I guess you set the whole map for the barrier area. Limit it to the area of red barrier glass only. It took ages until the barrier disappeared since the map is huge. The next time it crashed my game. This time you don't have to set the whole map for the barrier area since it's not a 4-teams-map.

HobbitPower (changes requested):
@sonicSMASH The team zones are still not correct. I get teleported back while walking in my team zone, but can cross the red barrier.
Check **all** team zones again. Edit the map and save it again with the latest CTF version. This will also add a new file (barriers.data), which is needed for the PR. So the testers can give feedback about the gameplay on the map.
P.S.: Quite a big map.

mrbesat2:
@sonicSMASH map and concept is solid, i like it, very big map, i can imagine alot of players playing here, team borders were fine, instead of needing blocks or needing to shift-jump up the hills, you can just use the water, although when i tried to test it twice in a row, i got a chestzone error and i couldn't move, maybe that could just be on my side ^^

HobbitPower (changes requested):
Team zones and chest zones now seem to work. 
A glass layer is missing at the top, your map is open to the sky. There are also gaps at the outer barriers (see picture).

Both sides have the same number of mese and diamond ores, but the iron ores are very different. I assume this was already generated before and is spread throughout the underground. 

About the map itself: It is interesting, albeit very large. I hope your speed/jump/gravity values will counteract the size of the map. I'm interested in more opinions on this and hope for feedback from the testers.
image

olliy1or:
review required label removed until the map is ready

sonicSMASH:
I added roof, adjusted iron ores, tweaked speed / gravity a tiny bit, patched barrier holes.

SoulS33ker:
Approved,no bugs detected,ores balanced.
will be interesting to see how such a big map plays out in-game.

HobbitPower (commented):
@sonicSMASH 
* You could edit the conf file (line 17-20) as there are unnecessary lines there. Nevertheless, it works.

* Big map, we'll have to wait and see how it performs in the game. Maybe we need to adjust the speed values etc. (1.3 might be enough). Hopefully lag and FPS will be fine.

* I'm not the biggest fan of adding blueberries to chests, Nade fight and Classes shouldn't become a second Classic mode. If you put a few blueberry bushes in the map, I think that would be better. I would rather advise against putting it in the chests and will discuss this with the second map manager.
P.S.: At least you chose blueberries instead of apples.

* Apart from that, the map seems to be good enough to be added. At least from my side

MoreKat (commented):
Cool map, very big, a lot of variety in terrain which is nice. I can see that you paid attention to detail, however I'm not sure that it will be fully appreciated since I doubt people will be using the underworld. I agree with HobbitPower about the blueberries, would remove them. I also would set the speed to 1.3 max. Finally I think you should add blocks to initial_stuff, at least enough to build up to the floating islands. With these changes I think the map will be good to go.

HobbitPower (approved):
Edit: The requested changes appear to have been made. However, there is now another problem.
Apparently there is a piece missing in the red barrier. @sonicSMASH 
image
And while you're at it, you might as well remove this. (Not the main problem but you have to edit the map anyway, so better fix this too)
image

Themostrandom:
My freind say he love your map

Themostrandom:
Your map seems ok on a functional level, despite all, it is very chaotic but I like the general style, I think adding a way to show on the central island would be beneficial. The presence of iron ingots is disturbing, perhaps adding a small percentage of metal would be good?

SoulS33ker:
tested,looks good to go :+1:

sonicSMASH:
Map is now called Terrarium. This was requested in case of a copyrighted name (unlikely but better to be safe).

HobbitPower (approved):
The map should now be ready for a merge.

MoreKat (approved):
Lovely whimsical map!

PR #298, opened 2023-07-18
Add new map 'Downtown Neighbourhood' by daveistaken

daveistaken:
screenshot
a simple map.


LuismiInTheZone:
Woa those are lots of commits

daveistaken:
> Woa those are lots of commits

Yeah, I just added a new branch on my fork because I made a second map and i cannot make a another after

src4026:
> > Woa those are lots of commits
> 
> Yeah, I just added a new branch on my fork because I made a second map and i cannot make a another after

Hmm. That gives me a good idea for an addition to the Map Maker README.md :P

daveistaken:
This pull will be drafted as I started editing the map and renaming it to 'The City of Blocks'

HobbitPower:
* At first glance, the map reminds me a little of the Mini Village Map (which has been disabled), but this one is much denser. Maybe too dense. With many players and a lot of block spamming, it could quickly become tiring to play there imo.

* The amount of ctf_map nodes is also quite high. It might be difficult to take the flag, as only the xpanes are breakable.

* Below the sewers (and the ores) there are several rows of ctf_map:stone. Why? Since you can't tunnel that way, it doesn't fulfil any purpose and the map just takes more time to load.

* For my part, I am currently rather against adding the map/ it has not the highest priority at the moment. Feedback from testers is still welcome.

daveistaken:
> * At first glance, the map reminds me a little of the Mini Village Map (which has been disabled), but this one is much denser. Maybe too dense. With many players and a lot of block spamming, it could quickly become tiring to play there imo.
> * The amount of ctf_map nodes is also quite high. It might be difficult to take the flag, as only the xpanes are breakable.
> * Below the sewers (and the ores) there are several rows of ctf_map:stone. Why? Since you can't tunnel that way, it doesn't fulfil any purpose and the map just takes more time to load.
> * For my part, I am currently rather against adding the map/ it has not the highest priority at the moment. Feedback from testers is still welcome.

@HobbitPower thanks for the feedback, however, I will be not maintaining this map, as this will have no changes (since I'm focusing on other ones).

HobbitPower:
All right, then I reject the map in its current form.

olliy1or:
HobbitPower made good judgements on the map, and since there will be no changes to these, I'll close the PR for now.

PR #302, opened 2023-07-31
Update Ant War

lazylier:
This is not ready yet. Any feedback would be nice. The changes so far are focused on improving navigation above ground.


MoreKat:
I think there needs to be a diagonal tunnel from orange to green and from purple to blue (I haven't checked out the changes yet so if you've added that already, disregard this message)

lazylier:
> I think there needs to be a diagonal tunnel from orange to green and from purple to blue (I haven't checked out the changes yet so if you've added that already, disregard this message)

I've already been thinking about that myself. If I do that I'll need to move the bases farther apart I think, or find some other way to make it not so short to tunnel. So far I've only made changes to the surface.

HobbitPower:
> >I've already been thinking about that myself. If I do that I'll need to move the bases farther apart I think, or find some other way to make it not so short to tunnel. 
I don't think it's necessary to move the bases farther apart and wouldn't recommend it. Adding diagonal tunnels won't make it to easy imho.

> >So far I've only made changes to the surface.

The changes to the surface are the addition of the butterflies + fireflies or is there something else? Perhaps you could list the changes you are making, that would be helpful.

lazylier:
Ok, I've made some big changes to the underground areas. I did move the flags, not just to make the new tunnels longer, but because I felt like the nearness of the flags contributed to people's confusion about where to go. There are also more mushrooms, and tunnels crossing the barrier in the center.

src4026:
The bases and the confusion are now cleared IMO. Feels easier on the eyes (might be only me) :joy:. Amazing work with the butterflies once again.
One thing I would like you to perhaps have a go if it works is something such as connecting those middle tunnels you have made with the ones along the perimeter of the bottom of the map or something that would encourage digging there as well, since as was mentioned by Kat previously that people get lost and unsure of what to do with the "digging part" and resort to the top's limitedness. The current tunnels seem to solve kind of solve the problem of the capturing but seems to make it quite easy if the team that has taken the flag blocks the bottom tunnels and people can't reach the other one back in time. I am positive with the current changes, and as per my guesses, the tunnels are one of the few parts left for you to reach a point of culmination in terms of balance.

lazylier:
> I've also heard from some players that there are too many flowers on the surface. So just in case you want to change something there.

I don't plan to change that. It's just how I wanted it to be. 

> The simplest solution in my opinion would have been to add diagonal tunnels on the lowest level of the map. Right where the wooden ladders end. I can't think of anything that would contradict that.

I don't know why, I just liked the way I did it more.

LoneWolfHT:
@lazylier I haven't tested my changes, I'll leave that to you. But you now have a solid structure for your idea that you can tweak

src4026:
The changes work for me. :+1:
As for the changes since I last reviewed, I still feel a need for a set of tunnels.

There could be a set of them that lead to the top of the map, but the exits to the surface are closer in relation to one another and they may be connected laterally to the teams on either side or diagonally opposite. These could be connected to the main exits as well if you'd like.
Or, there can be a set of tunnels like that but are a copy of the current diagonal ones but higher/lower, like HobbitPower's previous message. I feel there is a rising need for more ways to navigate diagonally, along with the additions in this PR. 
I also think that the graffiti feature suggestion by LandarVargan would be a great addition to this map especially in the future when it becomes implemented. I guess that's much further away :).

HobbitPower:
@lazylier 
After another test, I still think that it is better to add diagonal tunnels on the lowest level. I can only recommend this. Because I'm not sure if the previous changes are enough or if players will continue to complain.

If you want to continue relying only on the obsidian barriers, it might be better to adjust/expand the hint. Something along the lines of "The more times you try to capture the flags, the more paths become possible". (I'm sure that could be worded better). And depending on how it plays out in the game, we may need to increase the probability of opening the barriers later. It would only take a small change in the init.lua file then. So no big deal.

lazylier:
> After another test, I still think that it is better to add diagonal tunnels on the lowest level. I can only recommend this.

I'm not sure I understand what you mean by this. Do you mean the cross tunnels should connect to the main ladder instead of going up behind the flags? Why do you think that would be better? I'm guessing it might be to make going through the tunnels take longer. If so, the reason I didn't do that is because it would be trivial to tunnel up behind the flag anyway unless I add more indestructible areas that seem to really confuse people.

HobbitPower:
> I'm not sure I understand what you mean by this. Do you mean the cross tunnels should connect to the main ladder instead of going up behind the flags? Why do you think that would be better? I'm guessing it might be to make going through the tunnels take longer. If so, the reason I didn't do that is because it would be trivial to tunnel up behind the flag anyway unless I add more indestructible areas that seem to really confuse people.

I just meant that you could add two more tunnels at the bottom where the wooden ladders end. So far, you can already use tunnels there to get to the iron and then get from Orange to Purple, for example. But Orange & Green + Blue & Purple are not directly connected there yet. I think that would be good for the gameplay and I see no problem with adding multiple paths to the flags. 

However. Personally, I just find it difficult to assess whether the tunnels with the obsidian barriers are sufficient. Of course, we can leave it at that for now and wait again to see how the players react.

lazylier:
I've added connecting tunnels and put signs near the barriers to suggest that they can become breakable.

HobbitPower:
> I've added connecting tunnels and put signs near the barriers to suggest that they can become breakable.

In case you didn't know and in case you wrote something on the signs: You won't be able to read it in-game. 
I like the changes. I approve.

HobbitPower (approved):
It is better than before, but again we will have to wait for feedback from the players. Nevertheless, I approve it.

Info for the other map managers: The code of the init.lua file comes from LV, but he hadn't tested it yet afaik. So before the merge it might be better if he takes another look at it. And he would then also have to check whether it fits with the Barrier Area/map version.

lazylier:
> In case you didn't know and in case you wrote something on the signs: You won't be able to read it in-game.

I did not know that, I should remove them then.

lazylier:
I'm not completely sure about this, but I don't think I've had any better ideas yet. If it's good enough, I won't make any more changes, I hope.

HobbitPower:
@lazylier Would be better to edit it once more with the newest version of the CTF game (the version that you can find currently in the ContentDB should be enough). Right now there is no barrier area (not in the conf file like it used to be and there is also no barriers.data file, since it was not part of [this](https://github.com/MT-CTF/maps/commit/1875ff0ffcee3298f082f66a4a95db78c267d5a7) and following commits.).

Can't even test the map right now and we need to convert it to the newest map version anyways.

lazylier:
I've simplified the barriers to open when any flag is captured. Possibly the map would work fine without the lua stuff at all, which should be considered.

HobbitPower:
@lazylier I guess this needs another update, since there has been some changes. At least there are conflicts 
image

> I've simplified the barriers to open when any flag is captured. Possibly the map would work fine without the lua stuff at all, which should be considered.

I agree, it would work without the lua file. If you think it's too easy to capture, you can add some gravel in the tunnels. This would fit the design of the map and is not too easy to dig with the stone shovel.

lazylier:
Ok, I have no idea what needs to be done now.

MoreKat:
I'm sorry but I don't see how these changes are sufficient. There are still only two entrances, I don't think players are gonna dig up from the lowest tunnels. Players still have to squeeze through the ladders, where they are easily killed, to get to the flag. You seem worried that capturing is gonna be too easy while the opposite was true, and you still haven't made fundamental changes to fix that. I don't think this map is ready to be added. You must keep in mind that it is easier to make the map last longer when you feel like matches end too quickly than the other way around, for example by adding obstructions in the tunnels. This is why I don't think you have to be afraid to add more entrances and tunnels.

lazylier:
Thank you, that gives me some more ideas. It is a possibility to add another path to the lower tunnels, but more holes in the walls seems dubious. Given the size of the map, I don't know where I could add more tunnels that wouldn't significantly reduce the distance between flags. It was already easy enough to capture when the map was active. My perception was that the problem was players trying to tunnel straight to the other flags and otherwise getting lost from not understanding the map design. I could add one-way entrances from the surface.

HobbitPower:
> Given the size of the map, I don't know where I could add more tunnels that wouldn't significantly reduce the distance between flags. It was already easy enough to capture when the map was active.

I wouldn't say it was that easy to capture. Anyway, you could connect the two rooms (treasure chamber + mushroom chamber) to the surface. Connecting the other small rooms (where the steel ladder leads down) is also another option.
Or you could add new tunnels to the base. There are still unused walls.

lazylier:
Is this sufficient?

MoreKat (approved):
It is such a maze, I honestly think players will hate it lol. But I'm willing to give it another try. I hope you're willing to make changes again if necessary though.

PR #305, opened 2023-08-07
Add new map Silver Lands

LuismiInTheZone:
Continued from #282 , I couldn't have there this work, I needed to end it, and once again, I remodified all... Suggestions are always good, so don't wait to tell them, I want the best for this :D


HobbitPower (commented):
Overall, I don't see the map as an good addition to the server, as there are already better mountain maps. The terrain is very hilly, the design is very mixed up and in certain areas it doesn't fit in with the surroundings.
The map needs an update to the latest version of CTF anyway, and the map creator initially withdrew from CTF. I recommend to not to add the map.

LuismiInTheZone:
Ah well, I tried to do something different, a map that can be cool for snipers and rifles, but I guess you are right, I guess we can close it since I don’t see anymore a point to continue this since is very difficult to change all to something different. Thank you for your time 👍

LuismiInTheZone:
After thinking about it, I came to a conclusion that there will be much better maps than this one, and Hobbit said all. Thank you for everything guys, I'm closing this PR.

PR #270, opened 2022-12-20
new map: plain 3

farooqkz:
It's a 3 team map.

PR #307, opened 2023-08-09
Add new map: Plain glass

farooqkz:
Hello. Here's come my glass map idea. It's not ready yet and thus, no screenshot. But reviews are welcome. I also feel the map should be bigger and is small now. Am I feeling right? Do you think I should add more buildings in each team's zone?

Main map features:

 - Each team can see the other bases. Teams cannot veil their base.
 - Tunneling is possible only from the trees and river parts of the map.
 - Two buildings, one for ore and the other for storage. The storage room can also be used as a shooting base. That's why it has holes.

RFC @HobbitPower @gameit @LuismiInTheZone 

Edit: I can run a test server for this map on my SBC from home. Do you think it'll do any good?


farooqkz:
> Ok look:
> 
> The map is very small, so if there is playing lot of people is going to be very difficult to move around there. Also, if is possible, don't you think you made the map very big of underground? I think is not necessary that much, since no one is going to make a tunnel of 100 blocks underground, maybe with 20-40 blocks underground is ok if you want to do this type of map, apart that you made wrong the checkpoints of building before starting the match, since if I go to the trees and start going underground, when I'm only 4 blocks underground I get teleported back to base.
> 
> I don't really know what to tell you that can improve this, by the moment I think you should make the map bigger, maybe making the outer glass barrier lot of blocks away of where is now, and fix what gameit said, that grenades can explode to the other team through the wall.

Thanks for your comments. I've updated the map to make the distance between flags longer which hopefully solves the smallness problem. I also have made the underground less deep.

Regarding team zones, this is on purpose. First, teams can't tunnel till match has started, second, teams can't place their doors underground which enforces team's strategy.

Regarding grenades, according to LV/LW, grenades won't get through a 2 block wall.

farooqkz:
@src4026 For me the chests are spawning. However, I get the same error for flags as you. The map is small, yes. But I have lengthened the distance between flags. Is this enough? To get the into zone of some other team, you first have to travel to the corner of your zone and then travel to the corner of the other zone.

Regarding camping, What in this map encourages camping? Other than the storage room, of course, which encourages that in nade fight.

src4026:
> @src4026 For me the chests are spawning. However, I get the same error for flags as you. The map is small, yes. But I have lengthened the distance between flags. Is this enough? To get the into zone of some other team, you first have to travel to the corner of your zone and then travel to the corner of the other zone.

So try placing the flags again and fixing them please :+1:.

> Regarding camping, What in this map encourages camping? Other than the storage room, of course, which encourages that in nade fight.

If one knows the map and that you need to go from one corner of that map (in the water ponds right?), They could just sit there until someone comes and kill them, the same applies for all the teams. It would be better if there were other points of reaching the other sides other than the region of the pond. Also, I noticed the hint says something about trees, but I didn't check what it was. I will do so today if I can.

farooqkz:
Okay I've updated the map:

 - Flags are set correctly.
 - Fixed problem of chests. For now, no chest gets spawned in the storage rooms. CTF is complaining it can't even place 2 chests in each storage room and it's weird. Hints to solve it are very welcome.
 - I've added glass tunnels, in the hope that they'll solve both smallness and camping. The map is not big, yet. But the glass tunnels can be well defended. So I think unless there are too few players, the map should not end fast.

Problems not yet solved:

 - I think some buildings in each team's zone can make the map more fun. It feels too plain now. Any suggestions?
 - The problem of chests not spawning in the storage rooms must also be solved. According to my tests, not even 2 chests can be spawned in each storage room. I need help for solving this.
 - Not sure but is the map suitable for "Nade Fight"? I have doubts
 - Since I expanded the map, there might be still "void nodes" blocking the way for players which I've overlooked. If you noticed, let me know.

src4026:
I forgot to mention that the map was not entirely destructible when I first tested it. That is, the outer surface blocks of the cuboidal map. AFAIK the general rule of thumb is to make the map inescapable. Example:
image

farooqkz:
> I forgot to mention that the map was not entirely destructible when I first tested it. That is, the outer surface blocks of the cuboidal map. AFAIK the general rule of thumb is to make the map inescapable. Example: image

If you mean the pond/river and their inner area, that's on purpose for tunneling.

src4026:
> > I forgot to mention that the map was not entirely destructible when I first tested it. That is, the outer surface blocks of the cuboidal map. AFAIK the general rule of thumb is to make the map inescapable. Example: image
> 
> If you mean the pond/river and their inner area, that's on purpose for tunneling.

I mean the other surface blocks. In the case of the above image, I meant to show the blocks are the "walls" of the map that could be broken.

farooqkz:
Got it! Will fix it, God willing.

HobbitPower:
@farooqkz Same question as for #270 Are there plans for the map? Or can it be closed?

farooqkz:
Okay. I've fixed some stuff. Another review is welcome.

HobbitPower:
@farooqkz Because you asked. A short but not final review: 
* Small construction fail, I think it's self-explanatory: 
image

* Can you swap the position of the flags and the teamchests? I don't really want the base positions to overlap.
* Please check the large chest zone with the 120 chests. I can't open the chests there (but I can in the diamond blockhouses). The chests also spawn outside the map. So please re-edit this chest zone and make it smaller.
image

* Your team zones also overlap by one block, see: 
image

* What is meant by your hint? I think you're referring to the fact that you can tunnel underneath the trees? By the way, this is also possible in the "ponds". Please adjust the hint.
btw: the saplings will not grow as far as I know. So why did you put them there lol
* License and screenshot missing
* I won't say anything more about the map itself, I'll wait until you've fixed things. However, capturing can be very tedious. The concept is innovative, but I don't know if the players will like it. At least you can also tunnel. Otherwise I would reject the map immediately.

farooqkz:
> @farooqkz Since you've asked. A short but not final review:
> 
>     * Small construction fail, I think it's self-explanatory:
>       image
> 
>     * Can you swap the position of the flags and the teamchests? I don't really want the base positions to overlap.
> 
>     * Please check the large chest zone with the 120 chests. I can't open the chests there (but I can in the diamond blockhouses). The chests also spawn outside the map. So please re-edit this chest zone and make it smaller.
>       image
> 
>     * Your team zones also overlap by one block, see:
>       image
> 
>     * What is meant by your hint? I think you're referring to the fact that you can tunnel underneath the trees? By the way, this is also possible in the "ponds". Please adjust the hint.
>       btw: the saplings will not grow as far as I know. So why did you put them there lol
> 

Fixed these.

You sure saplings don't grow?

olliy1or:
when the map is finished, request a map manager to review

farooqkz:
Please do another review @olliy1or @MoreKat @HobbitPower

LuismiInTheZone:
@farooqkz didn't test it but I still see that screenshot is missing

AnonomousAccount:
Interesting idea, but somewhat of a conceptual fail. Funnelling players through small tunnels is pretty fun for PVP, but bad for actually "capturing the flag". 
Pretty much, it goes against this quote from the handbook, "The design of a map should encourage differing gameplay and tactics."

farooqkz:
> Interesting idea, but somewhat of a conceptual fail. Funnelling players through small tunnels is pretty fun for PVP, but bad for actually "capturing the flag". Pretty much, it goes against this quote from the handbook, "The design of a map should encourage differing gameplay and tactics."

Players can also tunnel.

AnonomousAccount:
Pr
oposed fixes:
All it takes is a couple damage cobble, spikes, or doors to basically invalidate one of the tunnels.
I would make the tunnels much larger (3x3-5x5) and shorter in length. I'd also add some sort of middle-area between each team, aswell as maybe a middle area for all the teams, something kinda like the attached picture
map

AnonomousAccount:
> > Interesting idea, but somewhat of a conceptual fail. Funnelling players through small tunnels is pretty fun for PVP, but bad for actually "capturing the flag". Pretty much, it goes against this quote from the handbook, "The design of a map should encourage differing gameplay and tactics."
> 
> Players can also tunnel.

This only adds one more strategy, and a very limited one, since there is only one tunnel route. Scout Class is basically useless in this map, and so are most guns.

farooqkz:
@AnonomousAccount Thank you so much for your account.

 - You are right. Tunnels are too easy to block. But it's not possible to "invalidate" them with doors or spikes Only with damage cobbles and other team zone neutral building blocks. I think making them larger(like 5x5) is a good idea. But I'm not modifying the map till I see other reviews as well.
 - Guns are useful when Scouts camp near their base and prevent enemies from closing in. Moreover, guns are useful if you manage to get from one side to the other side. But in tunnels, yeah the guns are useless.
 - About having a middle area between teams, I disagree unless players can go through(like with tunneling or something). The concept of the map meant to be like bases are too close, there is glass and you can see the other side but you can't do anything.
 - What do you think about adding glass tunnels from the ceiling from each zone to two neighboring zones or alternatively, all ceiling tunnels joining a common glass room above the map.

farooqkz:
> @farooqkz didn't test it but I still see that screenshot is missing

Thanks for notifying. But I think maybe it's better to add the screenshot after all modifications are made?

LuismiInTheZone:
Ok, I'll add my review, first errors, then personal opinion:

Mistakes:

- The barrier is not well placed, if I go to the glass walls that communicate with the other teams I teleport back to the base (it only happens with the green and orange team). Also, if I go to the lakes that are in the corners of the base, if I go under the first layer of grass, I will also teleport to my base.
- There are gaps that are not filled under the trees. The hint says to look in the pond and in the trees, but I don't understand what's there exactly.
- You used indestructible glass instead of outer barrier glass (textures are different)
- It is almost impossible to go through the tunnels you made with that many chests as AnonomusAccount said, but apart from that chest area is setted wrong, I see chests appearing outside the map. I also get an error that it cannot place all the chests in zone 1
- I still can throw grenades through the wall and I hit players. HobbitPower, gameit, s20 and I said this a long time ago, and I don't understand why you haven't corrected it yet...
- Screenshot still missing

Personal opinion:

The map is too small and it is almost impossible to capture the flag, and it's the same with making underground tunnels. From my point of view pvp is horrible, the spaces are very small and all the players are going to be fighting without even seeing their enemies. The decoration is honestly not very good, since it only has some giant diamond cubes and that's it, the rest can be done by taking some random terrain and putting it on the map. Sorry Shangul, but I honestly don't think this map will go very far if you don't change something.

LuismiInTheZone:
> > @farooqkz didn't test it but I still see that screenshot is missing
> 
> Thanks for notifying. But I think maybe it's better to add the screenshot after all modifications are made?

probably

LuismiInTheZone:
I just checked and saw another problem, I can't open chests. Can anyone else tell me if this happens to them too? If so I will notify LandarVargan

AnonomousAccount:
> I just checked and saw another problem, I can't open chests. Can anyone else tell me if this happens to them too? If so I will notify LandarVargan

Ye, chest opening is bugged for me, though, this could be because of the mispalced chests or smth

farooqkz:
Thank you @LuismiInTheZone and @AnonomousAccount for your reviews. I will keep this map as my todo list but for now I would like to concentrate on code rather than map which I'm good at.

HobbitPower:
@farooqkz So can we close this PR for the time being? 🤔

farooqkz:
> @farooqkz So can we close this PR for the time being? 🤔

If you think that way's better, sure.

farooqkz:
I think I'm entirely redoing the map based on Anonomous's suggestion, with some changes, later when I've got time for it. I'm interested in submitting this map to both JMA CTF and main CTF.

PR #309, opened 2023-08-13
Add New Map: Martian Aliens

165135:
Adding a map I've been working on that utilizes the init.lua. I know this is a new concept and I'm sure everyone will kindly find fault with everything I've done. :)

The init.lua does a bunch of things that LandarVargan and everyone else probably disagrees with:
- Registers a block (Without using new textures, using a pre-existing texture and modifying the color)
- Registers an item (Without using new textures, using a pre-existing texture without modification)
- Moves semi-large chunks of blocks (By deleting and then replacing them in a different location) (I understand the possibility of lag, testing req.)
- Manipulates velocity of players


a-blob:
The pull request is missing a screenshot, which is required for a map PR.
https://github.com/MT-CTF/capturetheflag/tree/master/mods/ctf/ctf_map#8-screenshot

165135:
> The pull request is missing a screenshot, which is required for a map PR. https://github.com/MT-CTF/capturetheflag/tree/master/mods/ctf/ctf_map#8-screenshot

@a-blob Sorry, I did.
Also that pr close was a mistake XD

sonicSMASH:
> Moves semi-large chunks of blocks (By deleting and then replacing them in a different location)
:P

martian_aliens/map.conf
```diff
@@ -0,0 +1,24 @@
...
+chests = return {{amount=15,pos1={x=1,y=11,z=85},pos2={x=102,y=16,z=169}},{amount=15,pos1={x=1,y=11,z=-1},pos2={x=102,y=16,z=83}},{amount=30,pos1={x=42,y=17,z=144},pos2={x=60,y=17,z=156}},{amount=30,pos1={x=42,y=17,z=14},pos2={x=60,y=17,z=26}}}
+phys_jump = 1.38
+phys_speed = 1
+size = return {x=102,y=58,z=171}
+map_version = 2
+enabled = true
+name = Martian Aliens
+phys_gravity = 1
+initial_stuff = return {"ctf_map:jetpack"}
+author = Peace12345
```
sonicSMASH, line 13: I would go with your in-game name

165135: I usually do ``author = Brain_Juice (aka Peace12345)`` or vice versa, forgot to ig

Customiphone7 (approved):

165135:
> What kind of approval was that? Something is fishy 🤔

Hmm, I have no clue :thinking: 
But what do you think of this abomination? :pleading_face:

src4026:
> @src4026 
> > What kind of approval was that? Something is fishy 🤔
> 
> Hmm, I have no clue :thinking: 
> But what do you think of this abomination? :pleading_face: 

I'll review maps later, yours is in my list :)

HobbitPower (commented):
A first review: 
The map is quite simple in design with flat terrain. The map is also quite dark, which of course suits the setting/theme of the map. It's good that you gave torches as an initial item, but whether that's enough is the question. Maybe you could place more mese postlights or something similar. 
I'm also not the biggest fan of covering the base with too many indestructible blocks.

Your code (init.lua) is what makes the map special. However, the beam doesn't work for me. I think it's supposed to let players move up, but it kills me instantly instead.
Also, it seems a bit buggy when the ufo "flies" upwards. This text also stays in the old place, unfortunately:
image
It could also be hard to capture the flags then, since the ufos are in the air and the indestructible glass makes it even harder. Of course, your "jetpack" comes in handy. But even if it is quite funny at the beginning, it can become quite annoying in the long run. It gives pillarers a huge advantage and I think it will annoy a lot of players in all type of fights as well. 

I will test it again later. That's it from my side for now, maybe more testers will play the map soon.

165135:
@HobbitPower 
I will admit, I am not good at making designs that look good, I'm more of a "make it functional" kind of person.
I'll think about either more initial torches, more mese lights, or adding lava to make everything brighter.
About the ind. blocks: Yeah, I was wondering about that. Fitting the theme of a UFO, I felt I had to make it see-through, maybe I'll add another custom block for "Re-enforced Glass" or smth.

The buggyness when the UFOs move upward could be just a visual loading side affect, or you're talking about just being glitchy overall, which I won't deny :/

The Flag Waypoints: I looked into moving them, but the current flag waypoint update function only sets the name, not position, so I would need to make a commit to the actual game for that, and I was trying to stick within a map-init.

Killbeam: Good question, I'll try some things.

HobbitPower:
@165135 I just wanted to check if there are any updates on the map. Are you planning to continue working on it in the future or rather not? No rush, just let me know when you have time.

HobbitPower:
I close this due to inactivity and no prospect of change in the near future. If there is interest in working on the PR again, it can be re-opened. Map maker has been informed some weeks ago.

PR #311, opened 2023-08-16
Edit Savanna Ravine barrier area for faster barrier removal

a-blob:


HobbitPower:
This was on my to do list as well, glad you fixed it.

PR #312, opened 2023-08-20
Add map: Giant Chess

lazylier:
I've tried to make this good. There are some hidden ladders, and most pieces have mese or diamonds somewhere inside.


sonicSMASH:
Pieces look nice, but overall the map looks boring. Maybe add tunnels if there already aren't. Also, you could replace "chess" with "Chessboard" in the title (my opinion :)

JackofLava:
lol, another player and I were working on a similar map.

JackofLava:
Also, does this map have two teams or 4 teams?

lazylier:
It's two teams. I'm still trying to think of what else could be added.

JackofLava:
In that case me and my friend will work on a four team chess map

MoreKat:
> In that case me and my friend will work on a four team chess map

If this map gets merged, your map will have to have more differences than simply be four teams instead of two teams. Originality/uniqueness is important. Just letting you know so it doesn't end up being a waste of time.

lazylier:
> lol, another player and I were working on a similar map.

I wouldn't want to interfere with a better map (if it is better). I didn't know anyone else was doing something similar. What is yours like?

JackofLava:
We are still in a designing phase, but I do know that we are using a whole different color pallet. Though we where also going to make the chess board and pieces large.

HobbitPower:
@lazylier An update to the actual CTF version would be nice. Also the licence is missing and you can delete the part behind `others` in the conf file.

This is no full review: 
The chess pieces look good. The different heights of the squares are also better than just having a flat board. Since I think the map will be very inviting for sniping, it at least offers opportunities for hiding. However, the different heights can also be annoying, as you always have to build your way upwards. Perhaps you could add some ladders?

As mentioned, tunnels might not be a bad idea, but at least the blocks are breakable. It just takes a while to dig through the different block types. 
I'm just afraid that the chess pieces can't be put to better use. They will probably mainly serve as sniper hiding places or when you pillar yourself up. Except in nade fight, where you can use and reach them better.

Speaking of nade fight mode: you can easily nade the base from the neighbouring pieces. Perhaps it would also be better to use two indestructible layers of blocks for the base area. To prevent nade spam from below. 
The base itself reminds me a bit of the tower map, but much smaller. To make the surrounds destructible is interesting, since you could create a floating base in the sky. Good thing the ladders are indestructible though. 

For my part, I will test the map in more detail at the next opportunity, but would be interested in feedback from other testers regarding the gameplay and whether the map can be fun.

Themostrandom:
Hello, for this map, I found a problem related to WorldEdit I think, a wall is under the map outside the area and I suggest adding entry in the counters, apart from that, the map seems correct.
screenshot_20240410_132210
screenshot_20240410_132117

HobbitPower:
@lazylier let me know when the issues are fixed. I think you should also add a 4.0 after your preferred license.

lazylier:
ok, I still want to add some more tunnels too

JackofLava:
Thanks for the feed back everyone and sorry for not responding sooner. My co-designer's and I have not been able to get back this project and we don't know when we will be able to, possible sometime during or after this upcoming Fall Semester for college. 

Again Thanks for all the advice, suggestions, and information, we greatly appreciate everything you guys are sending us!

lazylier:
Should be good now.

HobbitPower:
@lazylier I like the map, even if some people will dislike the differences in altitude. I prefer this to a simple flat chessboard, but I would like to see a few blocks as initial items. Ladders would be good. 
There are marginal differences in ore, but with the amount of iron, this is totally acceptable. The mese is the same on both sides. 

If you add the ladders (or other blocks), I will approve the map.

HobbitPower (approved):

MoreKat (approved):
The chess pieces look amazing considering the shapes you could work with. I like the variation in terrain and hope it will give for good gameplay.

PR #313, opened 2023-08-24
Fix water barriers

Emojigit:
This PR uses the new air and water variants of build-time barrier for maps (not working on disabled maps). After applying this patch, water flows of the maps will resume normal, and the load time will be faster.

~~This PR should not be merged before MT-CTF/capturetheflag#1218.~~ Can be merged right now.

Fixes MT-CTF/capturetheflag#1211


Emojigit:
All maps are tested one-by-one and all working.

a-blob:
@Emojigit  Wouldn't it be easier to just disable water checks for the old glass barriers and replace all the underwater ones with water barriers?
Also, what is changed in the map.conf files except the reordering?

LoneWolfHT:
If nothing should have changed in the map conf you can probably just remove those changes to be safe. My last commit fixed shadows not being enabled on some maps, which is why I included the changes

Emojigit:
> @Emojigit  Wouldn't it be easier to just disable water checks for the old glass barriers and replace all the underwater ones with water barriers?
> Also, what is changed in the map.conf files except the reordering?

LoneWolf have removed the water checks in the latest commits. 

And for the map conf, I did not modify the intentionally but included the changes in case the internal code changed them.

LoneWolfHT:
Yeah, I attempted to merge this but got sidetracked with work. It's probably too late now, as a few maps have been changed

Emojigit:
I will rework related PRs later, probably early November.

PR #317, opened 2023-09-08
Make it possible to get out of the pits without blocks in Water Academy

a-blob:
This PR makes a slice of the stone wall of the pit (the slice nearest to the flag) breakable so that players inside the pits with no blocks can dig out.


HobbitPower:
@a-blob I haven't worked with it much myself, but you could also replace the red barrier glass in the ocean with the new water barriers (blue barrier glass). If you want to do this, let me know.

a-blob:
> @a-blob I haven't worked with it much myself, but you could also replace the red barrier glass in the ocean with the new water barriers (blue barrier glass). If you want to do this, let me know.

Sure, I'll do that.

a-blob:
@HobbitPower Done, but now there's a conflict.

a-blob:
Conflict solved.

HobbitPower (approved):

PR #318, opened 2023-10-23
Make both water bodies the same depth in Lost Mountains

a-blob:
The map Lost Mountains has a body of water around each team. One is shallow while the other is deeper. This PR makes it fair by making both water bodies shallow. 
Now, all players who jump from the mountain will die, making it harder to speedrun.
The map's author Luis_Mi planned to do this, but he left.


LuismiInTheZone:
I would be delighted to help in any way if is necessary for this, any feature that makes it better is always needed. Thank you for doing this GreenBlob

HobbitPower:
Ngl the map looks better when the water isn't only 1 block deep. But it's certainly fairer and more equal now.
@LuismiInTheZone ik you are busy but if you want you can test the changes made to your map and share your opinion. Not necessary as the licence is CC BY-SA 4.0, just an offer.

HobbitPower (approved):
Approved, for reasons of fairness

Issue #319, opened 2023-11-23
Red barrier is only partially removed


HobbitPower:
The bottom layers of the red barrier are not removed at the start of the round when testing in a **local** world. This is not noticeable on maps whose lowest layers are filled with stone, for example, but on maps such as **Water Academy** and **The Stone Maze** (PR #299). This also occurred when creating a map myself, even after re-saving several times. Water Academy is added to the server and this does not occur in-game afaik. Nevertheless, it would be good to know whether new maps can be merged without hesitation.

A striking feature of all maps in map version 3 (including those currently on the server) is that the second (smaller) y-coordinate always has the value 16 instead of 0: 
![image](https://github.com/MT-CTF/maps/assets/76528136/dd2f2e55-2e5f-4fb0-9ecc-f87da96f7390)

LoneWolfHT:
It's safe to merge maps. I think the issue is fixed now too

PR #320, opened 2023-11-23
Add new map "Titanomachy: Aftermath"

HobbitPower:
NOTE: map was made with the newest CTF version (ContentDB December 2023). If you use an older CTF version, you will get an error message (because there were no indestructible walls). 

* When testing, the red barrier does not disappear fully, but this is most likely a general problem with the map editor. Nevertheless, this needs to be fixed. (#319) Edit: fixed ✅

* General feedback is of course appreciated, but especially on the placement and number of chests & ores. 

* The map has a small labyrinth/maze, but this is mostly for entertainment purposes. In my opinion, it does not simplify the capture of the flag too much, but feedback is welcome.


lazylier:
It looks amazing! The ores seem ok. I counted 10 diamonds and 76 mese for each team above ground, and a lot more in the big area beneath. It's not perfectly even I think but there's so much that it shouldn't matter. I'm just wondering how useful that underground area would actually be since it only has one entrance (or are there more?) What if you added one way entrances from each side?

HobbitPower:
> It looks amazing! The ores seem ok. I counted 10 diamonds and 76 mese for each team above ground, and a lot more in the big area beneath. It's not perfectly even I think but there's so much that it shouldn't matter. I'm just wondering how useful that underground area would actually be since it only has one entrance (or are there more?) What if you added one way entrances from each side?

Technically there are 5 entrances (2 waterfalls, 2 circular staircases and the main one). But I will consider it if I can find a way to make it fit. Thanks for testing!

omikekimo:
This map is absolutely stunning, every time I look around in it I see something new.  I have no idea how you made the wave in the water, genius. I am not going to bother with counting the ores underground because there is so much of it but above ground I counted the reds have 7 mese blocks while the blues have 8. Each team has 10 diamond ores above ground and 0 mese ore as far as I can see so thats fair and square. So apart from the mese block discrepancy (please someone else also check that) this map is perfect.

HobbitPower:
> This map is absolutely stunning, every time I look around in it I see something new. I have no idea how you made the wave in the water, genius. I am not going to bother with counting the ores underground because there is so much of it but above ground I counted the reds have 7 mese blocks while the blues have 8. Each team has 10 diamond ores above ground and 0 mese ore as far as I can see so thats fair and square. So apart from the mese block discrepancy (please someone else also check that) this map is perfect.

Both teams got 7 mese blocks, one is in the center, available for both teams. However red got 192 mese ores and 179 diamond ores (mostly underground, 8 diamond ores above). Blue got 194 mese ores and 175 diamond ores (also only 8 above the ground). I used //mix so there is a slightly difference between both teams. But since blue only got 2 more Mese ores and 4 fewer diamond ores, I see no need for a change. 

Thanks for testing it! I made the wave with the help of ctf_map:ignore btw.

SoulS33ker:
**MERGE ASAP**

MoreKat (approved):
This map looks incredible, the design is so intricate and visually pleasing. I think this maximizes the limited potential of the selection of blocks that we can use in CTF.

olliy1or:
I agree with Kat, the map looks awesome, my only complaint would be that there isn't a way to safely get to the lower level. While on blue side they'll survive the fall into the water. I might be missing something so please correct me if so

HobbitPower:
> I agree with Kat, the map looks awesome, my only complaint would be that there isn't a way to safely get to the lower level. While on blue side they'll survive the fall into the water. I might be missing something so please correct me if so

There are stairs leading to the underworld (under the tree), one spiral staircase per side around a lava fall and one waterfall per side :)

HobbitPower:
> I agree with Kat, the map looks awesome, my only complaint would be that there isn't a way to safely get to the lower level. While on blue side they'll survive the fall into the water. I might be missing something so please correct me if so

@olliy1or Are you talking about the underworld?
There is the main entrance available for both sides (under the tree/behind the skeleton hand). Then there are 2 waterfalls (one per side), where you can jump into the river. And 2 spiral stairways (one per side again) around a lava fall.

(If you are talking about the sea: you can survive the fall into water, no matter what side.) :)

olliy1or:
Ah, I missed a couple of entrances. Rechecked everything, can't wait to play on this map :)

PR #321, opened 2023-12-04
moon landing

TheEt1234:
map twist: the gravity is lowered from 1 to 0.16

in the middle of making this we realised that there was also a moon map so we tried to make things original with the lava and the white rockets

map license : CC0

by: frogTheSecond, Dark_Inferno

p.s. this is our first first time making a map


TheEt1234:
also i didn't know how to create a directory inside github editor without creating a file so that's what moon_landing_3/github editor trollery was

TheEt1234:
added screenshot... wow i should reconsider using github editor even for basic things

lazylier:
I found two problems with this map. First, the barriers don't disappear entirely, and it looks like some of the wall isn't made of red glass anyway.
barriers

Second, the bottom of the map isn't entirely indestructible. You can use worldedit to replace the destructible blocks there. 

Otherwise, it looks good.

TheEt1234:
oh yikes, okay will fix tomorrow

also i need to fix the fact that its too speedrunnable (possibly make the player slower+add more guns)

TheEt1234:
also i am unsure how i would fix the border since i did //replace air ctf_map:red_ind_glass or something like that

lazylier:
I think it's connected with *barrier_area* in map.conf

TheEt1234:
I didn't experience any barrier issues, works on my machine and i apologize for some of the bottom floor being destructable, my bad, there were also some worldedit skill issues i had to clean up

I also changed the movement speed to 0.5 to make speedrunning harder, i will be publishing the changes soon

lazylier:
Maybe it's just me, but the barrier problem is still there. Also, not really related to that, I do think you should change the area where the permafrost forms part of the barrier wall. It looks unlike the other terrain.

The speed change feels likely too slow. Maybe 0.75 something in between?

TheEt1234:
> It looks unlike the other terrain.
I will change it once a breaking issue is found since i don't really want to create a bunch of microcommits 
> The speed change feels likely too slow. Maybe 0.75 something in between?
Low gravity pvp is hard, that change was made to make it a lot easier and make the map less speedrunnable
0.75 i feel like is too fast for that

src4026:
> Maybe it's just me, but the barrier problem is still there.

With both of you on the latest version of CTF? Version 3.7 on CDB should have the commit for the barrier removal fixes.

TheEt1234:
The map was saved on a slightly older version

PR #322, opened 2023-12-15
Revise Volcano Map

lazylier:
This is my attempt at replacing the airships to make the Volcano map usable again. There's probably a lot that could be improved. I haven't touched the barriers, although they look kind of messy. I attempted to place the ships the same distance from their respective corners. Each ship has a tree (like the older model), as well as some hidden ores, which should be discoverable by observant players.


AnonomousAccount:
I really like your redesign of the ships, it captures a similar mystique to the original. The colours, however, are a bit overwhelming, I'd revert the balloons back to white wool, and maybe add some sort of coloured pattern instead. The deck of the ships is kinda empty too. IIRC, the deck on the original has a house on it, I'd suggest adding some sort of building. Overall, a good job for a draft 👍

olliy1or:
@lazylier when finished, request for review

lazylier:
This might need more but I don't have good ideas

Themostrandom01:
Add bots

MoreKat (commented):
Everything looks good, I think you did a great job redesigning the airships. However the barrier is kind of messy (see screenshots). There are parts missing and it leaves an imprint when it disappears. This seems to have already been there in ksandr's original map, I hope you can fix this. 
screenshot_20240608_223103
screenshot_20240608_222817
screenshot_20240608_222341

lazylier:
I've made some improvements to the barriers. I also replaced some ugly leftover stone on one side.

HobbitPower:
@lazylier It's still there. Would be good if you could add this with sand. 
image
The only downside to the map design could be that people spam nades below the flag. LV once said that there should be 2 layers of indestructible blocks to prevent this.
Apart from that it's ready to merge imo.

lazylier:
> The only downside to the map design could be that people spam nades below the flag. LV once said that there should be 2 layers of indestructible blocks to prevent this. 

Weren't nades changed to not damage through walls? I will try to fix the sand

HobbitPower (approved):

Themostrandom:
I have several suggestions for you: 
1. There are holes in the bottom of the volcano next to the ores. 
2. A block of silver sand is present in the volcano.
3. I don't understand the purpose of the earth blocks.
4. The lava under the water doesn't look very RP (you can't see the separation blocks, so it looks like a bug and isn't very pretty).
4. I'd suggest adding some kind of mine to visit the volcano's interior (as if flying ships were a civilization trying to understand how volcanoes work) and this would add access to minerals, which is very complicated and time-consuming.

Themostrandom:
(I just saw that @HobbitPower  suggested you put ctf_map:ignore at the map level so don't necessarily take that part of the message into account.)

lazylier:
> I have several suggestions for you:
> 
> 1. There are holes in the bottom of the volcano next to the ores.
> 2. A block of silver sand is present in the volcano.
> 3. I don't understand the purpose of the earth blocks.
> 4. The lava under the water doesn't look very RP (you can't see the separation blocks, so it looks like a bug and isn't very pretty).
> 5. I'd suggest adding some kind of mine to visit the volcano's interior (as if flying ships were a civilization trying to understand how volcanoes work) and this would add access to minerals, which is very complicated and time-consuming.

I'm not trying to redesign the whole map, so I don't know about making changes to the volcano or the ores. What do you mean by "earth blocks"?

Themostrandom:
> > I have several suggestions for you:
> > 
> > 1. There are holes in the bottom of the volcano next to the ores.
> > 2. A block of silver sand is present in the volcano.
> > 3. I don't understand the purpose of the earth blocks.
> > 4. The lava under the water doesn't look very RP (you can't see the separation blocks, so it looks like a bug and isn't very pretty).
> > 5. I'd suggest adding some kind of mine to visit the volcano's interior (as if flying ships were a civilization trying to understand how volcanoes work) and this would add access to minerals, which is very complicated and time-consuming.
> 
> I'm not trying to redesign the whole map, so I don't know about making changes to the volcano or the ores. What do you mean by "earth blocks"?

dirt block*

MoreKat (approved):

PR #234, opened 2021-10-23
Add "The Mines Of Vexillum" map

DragoniEU:
The `default:dirt` is just a placeholder, initial items are up for discussion.

The Mines Of Vexillum contains custom trees, mese and diamond ores, vast iron veins (thanks to KGM who coded them) and a couple coal ores, as well as lots of lava.

(It may be a little asymetrical, but it's still fair. I made sure that there are equal amounts of diamond and mese ores on both sides).

Enjoy!
screenshot_20211023_232633

PR #323, opened 2023-12-17
add new map: mines of vexillum

farooqkz:
Backlog: https://github.com/MT-CTF/maps/pull/234#issuecomment-1811089863


HobbitPower:
Screenshot missing (3:2 ratio)

farooqkz:
@HobbitPower fixed problem with barriers and included a screenshot

Themostrandom:
Hello, I have several comments:
- The poorly defined barrier between the two teams: 1 block too many on each side.
- There's air under some of the blocks, even if this is possible in a mine, I doubt it's deliberate on your part.
- The ores are poorly distributed between the 2 teams (iron, coal).
screenshot_20240410_133023
screenshot_20240410_133014
screenshot_20240410_133001
screenshot_20240410_132956
screenshot_20240410_132928

HobbitPower:
@farooqkz As TMR pointed out:
* Team zones need an update. You get teleported back while standing at the red barrier. 

* The ores are fine, there is just a huge difference for the coal. You could add some coal blocks, although there should be still enough coal for both sides.
* Replace the stone there with ind barrier glass: 
image
* This can be fixed too. Make sure to check the corner points of the maps then.
image

ClothierEdward:
The only thing I would change about this map would be adding grass to the outer edges. I find it weird that there's only grass in a few spots of the map.
Other than that, I like the idea and execution of the map. I would like to see this map be added to the server.

Themostrandom:
image
image
Hello, there are still some gaps in your map. The chest spawn areas are not very well placed (the chests spawn above the lava).

farooqkz:
closing this as I'm not interested on working on it anymore.

PR #324, opened 2024-02-02
Add new map "Mall Brawl" by oMikekiMo

HobbitPower:


src4026:
`screenshot.png` to be added? Really wanted to have a quick first look at the map when seeing the PR :)

HobbitPower:
> `screenshot.png` to be added? Really wanted to have a quick first look at the map when seeing the PR :)

yeah, he was already informed ^^ just didn't have time to update the PR earlier. Now the PR is complete.

HobbitPower:
A few words about the map (not a final review): 
* The mall itself looks good, but the positions of the bases irritate me a little. You can theoretically ignore the mall completely and walk around it. I fear that the mall could primarily only be used in building time to loot chests and ores and perhaps when capturing the flags of the diagonally opposite bases. But it also depends on the modes ig. In classes it could be used by scouts.  I would therefore wait for further feedback from other testers.
* The distance between the bases is also quite large. Between 255-275 blocks for the neighboring bases and 378 diagonally. There's no direct limit for this, but I personally find the distance a bit too large. We have to see whether the increased speed will counteract this.
* Orange and red are quite similar team colors. It would be better to replace one of the teams with purple.
* I haven't found any bugs so far, just little things that can be fixed. I don't know if this is what is wanted here (see picture, roadside). 
image
The same applies to missing blocks of earth under the red barrier, right next to the outer barrier. (Not a big deal.)
Also I don't know if it was intentional that the glass domes aren't connected to the building. But it's not a problem and it's easier to get into the mall from above. 
* Haven't checked the amount of ores. Will do that when I test the map again.

Just my first impression of the map, more feedback appreciated.

ScaleRaker:
Before I want to say anything about the map, I would like to say that I am not an official map tester, so there is no need to take the following comments into consideration. I would also like to add that I love this map. The unique look of the mall with it's levels and stores make this map amazing. There were only a few small bugs which I thought could be fixed.
1. Between the maps green and orange, the placement of the team chest changes. 
2. I don't know if this was intentional or not, but there may something wrong with the flooring here: 
screenshot_20240307_155438
3. In one of the cinemas, a block is missing (torches added for light): 
screenshot_20240307_155720
4. A few underground potholes: 
screenshot_20240307_160045
screenshot_20240307_160129
screenshot_20240307_160226
screenshot_20240307_160235
Overall, I love this map. As HobbitPower mentioned, I think the team placement could be better. Thanks for making this!

omikekimo:
Thank you guys for looking at this, I think you are  right about the team placement. My original thought was to put the teams in the delivery bays but I thought it might be too dark. Could maybe put them inside a shop, I don't know, I am open to ideas at this point. Looking at the defects in the flooring and roads you can tell I started rushing, thanks for pointing out the defects and potholes they all need sorting out. The gaps for the domes were intentional, they are based on the albert hall in london which has no connection between the dome and the building to allow for expansion/contraction. Maybe i will add some visually obvious support so it looks like that's how its meant to be. As for the red/orange thing, I am colourblind a little so i will take you at your word but is what you're saying about the colours a consensus view? i do see how the red and orange are similar but they don't seem as similar as purple/blue to me.

omikekimo:
I did everything except for the colour change, and things are soo much better as a result. 
flag
hobbit
scaleraker

Themostrandom:
Here is the list of problems/additions to be made or modified:
- Adding a forest or garden to empty areas
- Remove wooden signs as players can write on them
- Some double doors are badly made
screenshot_20240410_131141
screenshot_20240410_131213
screenshot_20240410_131108

MoreKat (commented):
Cool map, I really like the tile effect created by the combination of silver sandstone block and silver sandstone. Some remarks. First of all, the giant empty spaces around the mall (see screenshot) kind of bother me, the area is still part of the map because it is accessible but somehow feels like it isn't, I think it would already be a major improvement if you added some vegetation there. Also I noticed that the map is enabled for classic mode but doesn't have any ores, this is usually a requirement for maps enabled for Classic. I'm also not sure about the nade fight, it's quite a small map (as in the flags are not too far from each other) and there are a lot of smaller rooms so I'm not sure how well that'll work. We can give it a try of course, it's a quick fix if it causes problems.
screenshot_20240613_114715

HobbitPower:
> Cool map, I really like the tile effect created by the combination of silver sandstone block and silver sandstone. Some remarks. First of all, the giant empty spaces around the mall (see screenshot) kind of bother me, the area is still part of the map because it is accessible but somehow feels like it isn't, I think it would already be a major improvement if you added some vegetation there. Also I noticed that the map is enabled for classic mode but doesn't have any ores, this is usually a requirement for maps enabled for Classic. I'm also not sure about the nade fight, it's quite a small map (as in the flags are not too far from each other) and there are a lot of smaller rooms so I'm not sure how well that'll work. We can give it a try of course, it's a quick fix if it causes problems. screenshot_20240613_114715

The flags were located in these empty spaces in the first version. ^^

HobbitPower:
@omikekimo As Kat already pointed out, you should add ores if you want to enable the map for classic mode. :)

omikekimo:
I will be adding some extra bits to this map soon, I will go with the consensus and put some decoration outside so that it looks less bland, and some ores. I am just a bit busy at the moment so its hard to give designing this the attention that it deserves. Thank you all for looking at it for me, i will try to get it looked at within the next few weeks.

Themostrandom:
@HobbitPower for me you can delete the tag

ClothierEdward:
After playing on this map for a bit, I can confidentally say that I would really like to see this map make it to CTF. It's got the two things that would make me fall in love with a map instantly: interesting structures and being a four-team map. It's very fun to explore and has a lot of areas within it like the icerink, the cinema, the furniture shop, and much more.

There's only a few things I'd change with this map.
-First of all, the name. A minor nitpick but I'd prefer it if the name was simplified to "Sunshine Mall" instead of "Beige Sunshine Mall".
-When I booted up the map to test it, I got an error saying that the map doesn't support any of the three modes. Might wanna fix that.
-There's gaps in the hedges, specifically where the build-time barriers would be.
screenshot_20250211_202049
screenshot_20250211_202040
-In the underground tunnel, there aren't that many mese/diamond ores. Tried digging for a bit but there's only the ores you can immediately see. Please add more mese/diamond ores here. It'd make it more fun to play in classic.
screenshot_20250211_201620
-And last of all, there's this classroom with chairs and tables but the fenceposts that would make up the table are missing.
screenshot_20250211_203828

That's all I could find that needs fixing for this map. Fix these issues and it looks like we'll have a really good map on our hands!

Themostrandom:
image
image
Hi, 

I have 2 comments for your map:

The corridors may be too dark.
Grass would be missing in these 2 areas.


Your map seems complete to me. It might be speedrunnable but I think that with the density of chests, players will have enough stuff to defend themselves.

omikekimo:
Hi guys, thank you for the reviews, I did everything suggested except the name change because it is a clue. I did put more ores but I didn't put them in the car park but spread around the mall so that here is more of a reason to explore every corner. I don't know how to reproduce or fix any of the errors mentioned but I did make sure that all the modes were selected properly in the editor before I clicked 'finish editing' so if they are still there it is beyond me as to why. HobbitPower has been kind enough to upload all the relevant updated files so I hope that this is the final edit that will be necessary.

omikekimo:
It turns out I am able to reproduce the error. I still have no clue how to fix it though.

a-blob:
@omikekimo Error should be fixed now

a-blob:
Seems like a fun map overall. Didn't find any problems after testing everything. It looks amazing. Nade fight could be an issue since some of the teams are quite close to each other, making it easy to capture if you use the outside space, but I suppose that can be fixed easily if it causes problems. The lack of torches as an initial item could maybe cause problems in dark areas but I think generally there's enough lighting in the rooms and parts of the tunnels and the torches inside treasure chests could suffice.

a-blob (approved):

AnonomousAccount:
This map is awesome! One comment I have to add that **needs** to be fixed:

This map uses Protected Chests as decoration. While they look nice, if they are mined, placed down, and filled with an item, they result in an _indestructible block_ for other players.

a-blob:
I think maybe it might be better to fix that code-wise instead.

a-blob:
Fixed: https://github.com/MT-CTF/capturetheflag/pull/1412

AnonomousAccount:
Awesome! This map looks ready for merging then.

PR #325, opened 2024-03-01
Add Mars Map

AnonomousAccount:
Add a new map - A Colony on Mars?


ScaleRaker:
Before I state my assessment's results, I would like to say that I am not an official map tester and that you are by no means required to fix the following things. 
1. Underground pit. Happens all the time for me, just something that needs to be fixed :)
screenshot_20240306_145216
4. One of the doors is open 
screenshot_20240306_144456
5. The map seems a little barren. I fought with this same problem some time ago on my map. Adding some more texture to the landscape to make it feel a little more interesting could be nice.
6. The map's title contains a question mark. Any punctuation with the title of maps is not allowed. (I also figured this out the hard way)

Note: The overall architecture of the build and the subways are very nice. I appreciate all the work that went into it. Great map.

Themostrandom:
Here is the list of problems identified: 
- Remove iron signs: players can break them and write on them.
- Adding chests?
- Review the consistency of blocks: some are indestructible, others not[
screenshot_20240410_130244
screenshot_20240410_130243
screenshot_20240410_130227
screenshot_20240410_130224
screenshot_20240410_130222
screenshot_20240410_130221
screenshot_20240410_130220
screenshot_20240410_130156
screenshot_20240410_130154
screenshot_20240410_130136
screenshot_20240410_130101
](url)

Themostrandom:
Hello, your modifications are incorrect, the blocks reported in my previous message are still inconsistent, in the course of my tests, I realized that there were others. The chests are also better, but not as good: let me explain: you've added a zone with 16 chests:

_chests = return {{pos2={x=20,y=8,z=20},amount=16,pos1={x=10,y=8,z=10}},{pos2={x=116,y=9,z=192},amount=16,pos1={x=106,y=9,z=182}}} ._ 
screenshot_20240411_203530
screenshot_20240411_203443
screenshot_20240411_203436
screenshot_20240411_203423
screenshot_20240411_203421
screenshot_20240411_203420
screenshot_20240411_203419
screenshot_20240411_203418


But the old one is no longer present. It's better to make changes slowly and carefully than quickly and badly.

Themostrandom:
> The Indestructible blocks are the way they should be, and the chests should be good

I keep up the inconsistency of the blocks...

AnonomousAccount:
Map errors fixed, ready for manager review

SoulS33ker:
General bugs not detected✔ 
Approves inconsistency of indestructible blocks as anonomous style✔
**CRITISIM**
Map size 200x100
  personally i dont approve of narrow maps,makes it very hard to capture;take flags.
Plus spawn is at a corner of map,bounded by glass barriers on east & south,denies sneaking in from these sides.
Leaving tunneling the only method to take flag(now that's problematic),(even that would be some what difficult with map not having much depth)
**suggests:** improvements to spawn;
possibly a spawn design where those who are inside cant see outside(like ancient pyramids; see how lack of cover/open map is dealt with there,its easy to sneak it and take flag hence)

farooqkz:
The map is good and well polished overall IMO. One issue though: One team has a brown mushroom farm and the other team doesn't. There are plans to add mushroom farming to CTF and when that happens, the team with mushroom farming will be at an advantage.

farooqkz:
Additionally, if you make the glasses indestructible, it would be more fun.

AnonomousAccount:
> Additionally, if you make the glasses indestructible, it would be more fun.

@farooqkz you mean the glass of all the domes? Good idea

AnonomousAccount:
@HobbitPower @MoreKat @olliy1or Ready for manager review

PR #327, opened 2024-03-13
Sewer Under Village Update

winniepee:


winniepee:
I made this PR because I have to fix something from the map. Here's an example:
screenshot_20240305_230650
Here's the updated version:
screenshot_20240305_230844

HobbitPower (approved):

PR #328, opened 2024-03-19
Re-add Moon map

HobbitPower:
* Re-add the moon map based on these [poll results](https://discord.com/channels/447819017391046687/533752156235366418/1215999862605352960) 
* splitted the chest zone into two
* Revised authorship


farooqkz:
Congratulations!

PR #329, opened 2024-04-10
Add new map "Red vs blue in the stars" by Themostrandom01

Themostrandom:
Added a simple map, 2 bases, no elevation changes, no bulky blocks, 5 colors, no skybox just blue and redin the star!


HobbitPower:
@Themostrandom 
If you are using the latest map editor version, it should be `map_version = 3`. Have you used an older CTF version or perhaps edited the conf file manually?
This line should not be edited manually. It would probably be best to save the map again using the latest CTF version.
image

Themostrandom:
Yes, I had modified it manually, I've just modified it now.

PR #330, opened 2024-04-22
Add new map "Planet X" by Dr_Dinnerbone

HobbitPower:


Themostrandom:
List of problems in my opinion:

- The chests are too close together 
- there are caves with lava that are badly formed (lava that doesn't fill all of them)
- empty caves (not necessarily a problem)
- An open door
screenshot_20240426_172518
screenshot_20240426_172606
screenshot_20240426_172613
screenshot_20240426_172618
screenshot_20240426_172641
screenshot_20240426_172651
screenshot_20240426_172738

Apart from that, your map looks fine to me, although for my part, the snow bothers me (I don't like the sound and appearance in minetest). Personally, I don't think I'll enjoy playing on this map if no changes are made.

ScaleRaker:
Nice map! I personally like the new feel to a snowy world 😄 . However, a few things could be changed: The chest zones seem to have three on one side and one on the other:
screenshot_20240719_210740
This seems to constitute three chest zones while the other one is solely near the ice river. May be wrong about this, but that is what it appears to be for me personally. Additionally, there is kelp in the ice: 
screenshot_20240719_210812
Perfectly fine if this was intentional, but it is a little strange so I thought I might mention it.
Overall, the map is rather nice. I would personally like a few more buildings and vehicles, but that's probably just me.

ScaleRaker:
Also, probably not a problem, but one base is slightly higher than the other. Additionally, the hint could be capitalized with punctuation :)

HobbitPower:
> * An open door
>   screenshot_20240426_172738

Um, I would not consider this to be a major problem.

HobbitPower:
* Have you used TNT blocks for the map? 
image
* As pointed out by TMR and others: The chest zones are too small for the amount of chests. 
image
* Your hint is not necessary.
* screenshot must have a ratio of 3:2

PR #333, opened 2024-04-23
Add map "empty map" by Themostrandom01

Themostrandom:
An empty map to get back to Capture The Flag's roots: protecting your base


LuismiInTheZone:
I know I don't test anymore maps, but I don't think a map that is completely empty is going to be good for the game

LuismiInTheZone:
> An Empty Map Would Be Nice Since That Will mean Players Have To Work Together On A Strategy, Not Just Using The Geometry Of The Map To Their Advantage

The snipers, rifles, all guns are going to dominate the match, and also a map is meant for having a new experience of decoration, movement and pvp, a new adventure of trying to escape with the flag… for this you can just go to some normal pvp server where there are only blocks and swords… Also in this map will probably win the team with better skill and weapons.

LuismiInTheZone:
Also I don’t think a map where you can find buckets of lava and water are going to enter in game

PR #334, opened 2024-04-23
Add new map "Cargo Port" by Themostrandom01

Themostrandom:
I'm putting this PR back on my account, thanks to hobbitpower for making the original one for me.


HobbitPower:
Moved from #289

Themostrandom:
@LuismiInTheZone

Everything looks all right, but the only problem (for me) is that the map is too small and easy to go to the other base.
I also keep thinking that where's the base put... that looks like the map abandoned battlefield... But in general I like a lot how you made the buildings, looks very cool but... I don't know, I feel something is missing... 🤔

Now to the .conf file part:

Why not adding classic to your map? I don't know if you have some variation or problem that can affect to classic...

Keep in there, great job

------------------------------------------------------------------------------------------------------------------
@lazylier

It's nicely built, but maybe a little small. It's confusing that the message saying to look in the sewers, but there doesn't seem to be anything there to find.

---------------------------------------------------------------------------------------------------------------------
@a-blob

I like the builds on the map. I don't really see a very big problem with the map being small. However, I do find it weird that the cargos are made out of wool blocks. It looks unnatural.
I'm not sure what the hint means because I didn't find anything in the sewers. I find the map a bit like Abandoned Battlefield.
I didn't find any problems with the map for now.

---------------------------------------------------------------------------------------------------------
@HobbitPower

Will add the WIP Label, since Terraformer was about to change/add some things.
(@a-blob I understand what you mean about the wool, but Minetest unfortunately only offers a small selection of red/blue blocks)


---------------------------------------------------------------------------------------------------------------------
@src4026

So I had a look at the map. Everything seems well and good for now 👍. I found some inconsistencies in the block types, that is, indestructible and destructible variants of blocks as in the video below 👇
untitled.mp4

And I would like to mention a block that might have been placed on accident in the ship:
Screenshot from 2023-06-17 16-19-59

The current hint doesn't make sense as far as the treasures go, as they are located in the ship and not the sewers (Or are treasures in the sewers too?). [as noted by other people who reviewed this map too]
Overall, the map is good, and well-made. The sewer system was quite good in design aspects, mine was just a lazy one, I suppose I should not put too much burden on myself as my map had other things more important. I haven't checked everything for your information.

--------------------------------------------------------------------------------------------------------------------
@src4026

I have edited the above comment as there were some issues in the images ☝️

----------------------------------------------------------------------------------------------------------------------
@LuismiInTheZone

Ok the map looks great, maybe some indestructible bugs, I don't know if you did it like that on purpose... (other indestructible bugs are leaves)
errors
In the image, at the point where there are leaves, from there, I can't destroy them anymore.

Also the hint doesn't make sense, I would recommend you puting something about the ores.

(I don't know if you noticed this but mese is indestructible)
@[LuismiInTheZone](https://github.com/LuismiInTheZone)
Contributor
LuismiInTheZone commented [Jun 17, 2023](https://github.com/MT-CTF/maps/pull/289#issuecomment-1595769335)

Another thing, ores are wrong, there aren't the same ammount in each team part:

Team red: 4 diamonds, 10 mese (indestructible)
Team blue: 11 diamonds, 6 mese (indestructible)

Then you also have the ones of the middle, but those can be taken by both teams

-------------------------------------------------------------------------------------------------------------------
@LuismiInTheZone

You changed almost nothing of what we told you, you only changed the trucks, which are now indestructible, and the minerals, but they are still unbalanced:

red: mese 21, diamonds 10
blue: mese 19, diamonds 14

Please read all of our comments again and correct them, and later, keep the minerals balanced, delete them and put them again correctly if you want.

--------------------------------------------------------------------------------------------
@LuismiInTheZone

Ok here is the review: the ores are now correct, the problem is that you didn't fix what we told you of destructible/indestructible blocks. Isn't necessary but would be better if you do it as we told you to do everything destructible/indestructible. You can find all our comments again here. Another thing, this is for my opinion: if I was you, I would put a steel pickaxe, because if not maybe is a bit difficult to get all those (my opinion, maybe isn't necessary since there are a lot of chests. If someone more can say what he thinks of this wouldn't be bad too). But nice, keep there.

------------------------------------------------------------------------------------------------------------------
@LuismiInTheZone

And... what did you change?
-----------------------------------------------------------------------------------------------------------------------------------
@LuismiInTheZone

There are still some parts from the truck that are destructible, but for the rest, seems is ok

----------------------------------------------------------------------------------------------------------------------------
@src4026

opinion, maybe isn't necessary since there are a lot of chests. If someone more can say what he thinks of this wouldn't be bad too).

I think the current stone pickaxe is fine. Also, I'll review the map sometime later. 👍

-----------------------------------------------------------------------------------------------------------------------
@ScaleRaker

I definitely liked this map when I tried it out. The only things that I thought was weird was the fact that some parts of the trucks were indestructible while other parts were destructible. Also, the inside of the boat is kind of plain. Additionally, the map could be four-team. I don't know what HobbitPower would say about this, since that would make the map even closer to Abandoned battlefield, but yeah, I feel like that might be good. So, yeah, the map is definitely very nice!
P.S. The hint in README should not have a space between the exclamation mark and "boat".

--------------------------------------------------------------------------------------------------------------------------
@HobbitPower

    Additionally, the map could be four-team. I don't know what HobbitPower would say about this, since that would make the map even closer to Abandoned battlefield, but yeah, I feel like that might be good.

The way the map is set up, I would rather stick with a 2-team map.
As for the similarities to ABF: there are some, but I would say that the water in the middle, for example, makes a difference. I'm currently neutral towards the map.

---------------------------------------------------------------------------------------------------------------
@ScaleRaker

Hmmm, ok.
-----------------------------------------------------------------------------------------------------------------------
@Themostrandom01

When will the map be in game?
--------------------------------------------------------------------------------------------------------------------
@MoreKat

    Friendly suggestion to capitalise 'port' too in the title, so 'Cargo Port'
    Pick a better screenshot. Your screenshot should give a clearer overview of the map. Take it from one of the corners with jungle so both base buildings are visible.
    I think it will be too hard to capture with the building in which the flag is situated being completely unbreakable. I suggest either making it breakable or adding some more entrances
    I suggest adding more torches to initial_stuff. The skybox makes the map quite dark.
    I think another good addition to initial_stuff would be ladders.
    Maybe consider making more things breakable, for example the structures in wool (not the cranes). In my experience players prefer a breakable map over a map in which everything is unbreakable.
    Overall I like the design of the map. The cranes and trucks look really good. The map looks aesthetically cohesive. There seem to be no bugs.

---------------------------------------------------------------------------------------------------------------
@Themostrandom

I made all the requests, I let you test it, I added sewers under the bases for an additional access, doors and I made the containers breakable.

MoreKat (changes requested):
In my last review I told you to add more entrances, I don't really see this as having been accomplished. You've got a big entrance in the front, true, but the rest is still unbreakable. I suggest adding at least a small back entrance OR making the roof breakable (I prefer this last option). Otherwise I think capturing will be too hard, the players also have to cross a giant canal and travel quite some distance. Also please make sure your screenshot has a ratio of 3:2 (currently it's 1920x1017 = ~1.9 (instead of 1.5)) AND take your screenshot without texture pack enabled (the mese lamps look blue in your picture, I assume this is a texture pack).

HobbitPower:
@Themostrandom The screenshot is still wrong. It should show the entire map + without the red barrier + the name has to be `screenshot.png`

Themostrandom:
> @Themostrandom The screenshot is still wrong. It should show the entire map + without the red barrier + the name has to be `screenshot.png`

I did, it seems my screenshot was already correct (in size), I did the same thing again by removing the red barrier.

Themostrandom:
@MoreKat I did what you asked. I am available if needed to make other changes on my map.

cargo_port/map.conf
```diff
@@ -0,0 +1,21 @@
...
+teams = local _={};_[1]="flag_pos";_[2]="enabled";return {red={pos2={z=242,x=0,y=0},[_[1]]={z=226,x=15,y=24},[_[2]]=true,pos1={z=121,x=145,y=66}},blue={pos2={z=242,x=0,y=0},[_[1]]={z=16,x=130,y=24},[_[2]]=true,pos1={z=0,x=145,y=66}}}
+chests = return {{pos2={z=241,x=144,y=24},amount=100,pos1={z=149,x=1,y=24}},{pos2={z=93,x=144,y=24},amount=100,pos1={z=1,x=1,y=24}}}
+phys_jump = 1.25
+phys_speed = 1.25
+size = return {z=242,x=145,y=66}
+map_version = 3
+enabled = true
+name = Cargo Port
+phys_gravity = 1
+initial_stuff = return {"default:cobble 40\"","\"default:pick_steel\"","\"default:torch 15\"","\"default:ladder_wood 40"}
```
HobbitPower, line 12: use this for the line: `return {"default:cobble 40","default:pick_steel","default:torch 15","default:ladder_wood 40"}`
I hate an error while testing.

HobbitPower (changes requested):

cargo_port/map.conf
Themostrandom: > use this for the line: `return {"default:cobble 40","default:pick_steel","default:torch 15","default:ladder_wood 40"}` I had an error while testing.

done :)

HobbitPower (approved):
I think it's good enough to be merged now. Provided the second map manager doesn't find any more errors.

MoreKat (approved):

PR #336, opened 2024-04-25
Pvp arena

Themostrandom:


SoulS33ker:
**PR files**
1.conf ✔
2.mts✔
3.screenshot 3:2 ✘
**MAP**
1.no errors ✔
2.flag area ✔
3.team chests ✔
4.build time barriers ✔
5.treasure chests ✔
6.flag capture-able ✔
**MECHANICS**
1.equal chance of winning ✔
2.base position ✔
3.terrain balance ✔ 
4.ore count ✔  (M:12,D:15,I:15,C:18 per side)
**CRITICISM**
Map too small,100x100 detected,speed run-able.

HobbitPower:
@Themostrandom SoulSeeker is right: screenshot must have a ratio of 3:2. also take the picture **without** the red barrier.

SoulS33ker:
a definite improvement :) well done,no bugs
screenshot ratio seems about right,someone with script check it

Themostrandom:
> a definite improvement :) well done,no bugs screenshot ratio seems about right,someone with script check it

thx for your review 👌😊

ClothierEdward:
The map is pretty fun to play on and it's looking quite solid so far. The block palette is really cool with how you used wool and silver sandstone. I also like how in the middle, there are barricades setup that you can use to block attacks. 
However, the map isn't perfect. There are quite a few things that do need fixing:
1. The first wall is too hard to scale for both enemies and teammates. I'm not a fan of how the first wall on each side is pretty much unbreakable. It makes it impossible to back into your base if you don't have any blocks. I would suggest making the first wall on each side a little thicker with a layer of mineable blocks while leaving the rest as unbreakable. 
2. It would be cool to be able to break the barricades in the middle of the map. It could serve as an additional source of blocks and you could use them the rearrange them into even stronger barricades.
3. The two rooms at mid are really empty for some reason. Maybe add staircases there or move the chests into those rooms.
4. Make the map longer so it is harder to speedrun. As of right now, it is really easy to speedrun this map in under a minute which some people may not like. Making the flags more than 150 blocks apart from each other should do the trick.
Make as many of these changes as you can and this already pretty solid map will certainly be even better.

Themostrandom:
> The map is pretty fun to play on and it's looking quite solid so far. The block palette is really cool with how you used wool and silver sandstone. I also like how in the middle, there are barricades setup that you can use to block attacks. However, the map isn't perfect. There are quite a few things that do need fixing:
> 
>     1. The first wall is too hard to scale for both enemies and teammates. I'm not a fan of how the first wall on each side is pretty much unbreakable. It makes it impossible to back into your base if you don't have any blocks. I would suggest making the first wall on each side a little thicker with a layer of mineable blocks while leaving the rest as unbreakable.
> 
>     2. It would be cool to be able to break the barricades in the middle of the map. It could serve as an additional source of blocks and you could use them the rearrange them into even stronger barricades.
> 
>     3. The two rooms at mid are really empty for some reason. Maybe add staircases there or move the chests into those rooms.
> 
>     4. Make the map longer so it is harder to speedrun. As of right now, it is really easy to speedrun this map in under a minute which some people may not like. Making the flags more than 150 blocks apart from each other should do the trick.
>        Make as many of these changes as you can and this already pretty solid map will certainly be even better.

I'm going to do that, but I'm keeping the old version in case not everyone likes the changes I've made.

Themostrandom:
I've added 2 entrances and ladders to the map, as well as a skybox.

PR #343, opened 2024-06-26
Add more chest on bridge map 

Themostrandom:
- [ x ] This PR has been tested locally
- Suggestion : https://discord.com/channels/447819017391046687/1008040841102766153/1253879184233730058


LuismiInTheZone:
What was the result of the suggestion for curiosity, that I’m not currently at the discord

AnonomousAccount:
> Out of curiosity, what was the result of the suggestion? that I’m not currently at the CTF discord

pretty positive, 12-3, though players always like when there are more chests

HobbitPower:
If a PR, then it should be divided into two chest zones (30 chest per side).

Themostrandom:
I allowed myself several things: remove the excess map (outside the map surface) put 30 chests per zone and make sure they don't appear in the water (I'm not sure about that).

HobbitPower:
You removed the modes for the map. Also:
image

Themostrandom:
> You removed the modes for the map. Also: image

I don't have this :/

Themostrandom:
@HobbitPower its okay now
I fix bug and add ctf_map:ind_water and ctf_map:ind_stone_red

HobbitPower:
Everything seems to be working. The screenshot is a bit outdated now, but I don't think we need to adjust that. ^^
Approved.

Themostrandom:
> Everything seems to be working. The screenshot is a bit outdated now, but I don't think we need to adjust that. ^^ Approved.

That realy need an other approval ?

HobbitPower (approved):

PR #345, opened 2024-07-11
Add new map "Peaceful Battles"

SaiPanneerselvam:
I would like to add a new map named Peaceful Battles, thank ye.


HobbitPower:
@SaiPanneerselvam Haven't tested it yet, but: 
* you have no modes selected
* take a screenshot of the map without the red barriers (so after the round started. speaking of the red barriers: are there holes or is it just not fully loaded?
* the hint is useless btw. You don't have to add one, but if: it should give more information.

LuismiInTheZone:
Quick test: 

I can't start the map because when build time ends the game crashes, I don't know what you did to the map
- All that said HobbitPower before me
- The placement of the barriers and zone areas are wrong, I keep teleporting back to the flag and can't move.
- Red Barrier building is also wrong, there are parts of barrier missing through it
- Top ceiling and two of the side of parts glass barrier are missing
- Chests placements are also wrong
- Did you put chests there? Eliminate all the chests you put as they won't work.
- Holes underground that would be recommended to fill up
- You need to make the bottom layer indestructible. 

Honestly the decoration is a bit common and this is like any other map without anything different, so I don't know if it would be added, anyway if you want to continue with this fix everything. In case you need help, you can find information in the readme file, the CTF handbook: https://ctf-handbooks.github.io/starter-topic.html or join Discord (in case you are not already)

SaiPanneerselvam:
ok then thanks for suggestions but I might just quit on this map and create a different map altogether.

SaiPanneerselvam:
Ya know what? I might actually try to rebirth this pr. Probably as a new pr though. I'll see if I can get to it!

PR #346, opened 2024-07-18
Update "abandoned_isles"

SoulS33ker:
increases water height,adds ladders,reverts to old desert theme.
iron ore count changed : 70:70
note: treasure chests are meant to be on water surface.....~~abandoned isles is a magical place~~


LuismiInTheZone:
I don't know what to tell you, everything is fine and personally I like the rework that you've made, but... Since I don't know what's the exact criteria that map managers follow, I'm afraid that it might have some problems since you just mixed both older versions and added water, but as I said, I personally like it.

HobbitPower:
It's a shame that there are no infinite river water sources :/ @SoulS33ker

SoulS33ker:
nothing new,just fixed a waterlily in air ,added few blocks here & there.

MoreKat (commented):
It looks good and I will be glad to have this map back. Only remarks that I have:

1. Fix the spelling of the hint: pillers = pillars
2. Why the name change? I would prefer keeping the name 'Abandoned Isles' (without the 'revamped', that's stupid)
3.  I suggest to make the (iron) ores (70 on each side) more accessible and adding a hint about their whereabouts. This is not a must but I would like to hear other thoughts about this.

SoulS33ker:
99% of ore locations are as left by original makers,i just balanced both sides,i dont mind changing them.

SoulS33ker:
iron ore now pops out on surface to hint more ores below below where ever available

MoreKat (approved):
Hopefully this review will get a thumbs up from @SoulS33ker ??

HobbitPower (approved):

PR #348, opened 2024-07-19
 2 bases, 2 teams, let's pvp !

LeoLeg2009:
Test my new map ;)


LuismiInTheZone:
First of all, I would need you to rename the folder and put low bar ( _ ) between words, for example, put 2bases_2teams_lets_pvp, but I would recommend shortening the name of the folder and map (also it's best not to add accents or commas in the folder name).

Problems:

- Ceiling glass of the map is missing
- Please remove the chests as it will not work and may cause problems (I don't know if furnaces are needed to be deleted too? @HobbitPower ?)
- Please fix this, feels weird and I think it's better to leave everything with barrier glass 
b
- Please make at least the 5x5 of the floor of the flag indestructible.
c
- Chests seem to have problems spawning, try splitting the chest spawn into two (for each team) and removing the chests you placed directly without spawning.

Opinion:
For the rest seems ok, the houses you built are really good and red barrier and limits of build seem ok. For decoration the map is very empty, you are going to need to build something else or do something different if you want the map to be added, because if not I don't think this can enter into official CTF, since it's very simple to what new maps have.

LuismiInTheZone:
Also please fix the screenshot, it needs to be 750x500 and 2:3 and needs to be of all the map, not just one house.

PR #351, opened 2024-07-19
2 bases, 2 teams, let's pvp !

leolegPremiumPlus:
I think I finished…


AnonomousAccount:
Ok time for some initial issues: 
-You spawn players with a bronze sword which is not reccomended. In classes, this would give medics and scouts a sword. 
- Diamond pick and shovel is wierd, but whatever
- 99 torches is alot, i'd suggest reducing it to 10-30, as the map does not have any places that need lighting up (aside from at night)
- I would suggest making two chest zones instead of one. Making a seperate chest zone for each team gives the same number of chests to each team
- 100 chests is alot, you should lower it to 10-30 per team (20-60 in total)
- The name is pretty basic, you should think of something more unique and memorable

AnonomousAccount:
Game play comments: 
From just looking at the screen shot, the map is a bit too basic. It could do with adding some more structures and places to hide. It's also a bit small and i would bet it would be a short map. The diamond pick also means that someone could easily make a tunnel and be at the enemy base in less than a minutes

LeoLeg2009:
> Game play comments: From just looking at the screen shot, the map is a bit too basic. It could do with adding some more structures and places to hide. It's also a bit small and i would bet it would be a short map. The diamond pick also means that someone could easily make a tunnel and be at the enemy base in less than a minutes

What about Coast of Nostalgia, Plains of the Past and the other non builded maps ?

LuismiInTheZone:
The map is really nice, but I think if you added more structures it would be even better, so you will have more chances to include the map in the official CTF (and structures that are not copied from the wall, some new designs), also btw orange team have 3 towers and red 2. You say in the comment about Coast of nostalgia and all those maps because not having buildings, those were the first maps ever made in CTF, and were made by the first owner of CTF rubenwardy.

The only problem I see is that there are too many chests, and that's why (I guess) some chests spawn in the air. I would recommend splitting the spawn (as I said in the last review) and putting 30 in each spawn. Anono said to remove the shovel but that doesn't matter much, you can keep it if you want, it won't affect the game. I also said before that I need you to change the name, because it's too long, and that the folder needs to go with _ (2bases_2teams...)

LuismiInTheZone:
Screenshot still not 750x500, please fix it

LeoLeg2009:
> The map is really nice, but I think if you added more structures it would be even better, so you will have more chances to include the map in the official CTF (and structures that are not copied from the wall, some new designs), also btw orange team have 3 towers and red 2. You say in the comment about Coast of nostalgia and all those maps because not having buildings, those were the first maps ever made in CTF, and were made by the first owner of CTF rubenwardy.
> 
> The only problem I see is that there are too many chests, and that's why (I guess) some chests spawn in the air. I would recommend splitting the spawn (as I said in the last review) and putting 30 in each spawn. I also said before that I need you to change the name, because it's too long, and that the folder needs to go with _ (2bases_2teams...)

The screenshot is 700:500 and there are 42 chests and no flying chests

AnonomousAccount:
> > Game play comments: From just looking at the screen shot, the map is a bit too basic. It could do with adding some more structures and places to hide. It's also a bit small and i would bet it would be a short map. The diamond pick also means that someone could easily make a tunnel and be at the enemy base in less than a minutes
> 
> What about Coast of Nostalgia, Plains of the Past and the other non builded maps ?

These maps use interesting terrain for intresting gameplay

LuismiInTheZone:
> > > Game play comments: From just looking at the screen shot, the map is a bit too basic. It could do with adding some more structures and places to hide. It's also a bit small and i would bet it would be a short map. The diamond pick also means that someone could easily make a tunnel and be at the enemy base in less than a minutes
> > 
> > 
> > What about Coast of Nostalgia, Plains of the Past and the other non builded maps ?
> 
> These maps use interesting terrain for intresting gameplay

Not really, it’s because were the first maps

ScaleRaker:
This map has a very nice base :D. However, like all PRs, I have a few criticisms for this one:
1. The bases are not in the center on the X axis. The path, which runs through the middle of the map, is offset by 10+ blocks.
2. The team colours might be too similar (Check with HobbitPower for more info about this)
3. A strange open space (torches added to show what I mean) below the base: 
screenshot_20240720_164328
4. This gravel which strangely shows up in some and doesn't in others: 
screenshot_20240720_164244
5. There seems to be more ore in one base than the other (red has two layers of ore, orange has three)
6. As others have said, the length could be longer from one base to the other
7. As others have said, some more unique structures could be nice
None of these things have to be done. I'm just a noobie map tester, so take my criticism with a grain of salt :)

LeoLeg2009:
> This map has a very nice base :D. However, like all PRs, I have a few criticisms for this one:
> 
>     1. The bases are not in the center on the X axis. The path, which runs through the middle of the map, is offset by 10+ blocks.
> 
>     2. The team colours might be too similar (Check with HobbitPower for more info about this)
> 
>     3. A strange open space (torches added to show what I mean) below the base:
>        screenshot_20240720_164328
> 
>     4. This gravel which strangely shows up in some and doesn't in others:
>        screenshot_20240720_164244
> 
>     5. There seems to be more ore in one base than the other (red has two layers of ore, orange has three)
> 
>     6. As others have said, the length could be longer from one base to the other
> 
>     7. As others have said, some more unique structures could be nice
>        None of these things have to be done. I'm just a noobie map tester, so take my criticism with a grain of salt :)

I can't build bc im not in my grandparents house with the macos that have 5.8 MT that I can build the map, so I can't but can you help me pls ? I'm not good at building and for the teams i can change the orange team to blue team in the .conf but the towers…

LeoLeg2009:
So if you want to help, you can

ScaleRaker:
I can DM you about it.

HobbitPower:
Licence is missing btw

PR #352, opened 2024-07-21
Add new map "Abandoned City" by SG

ChandlerBing1994:
Two camps in one city. Many buildings are connected by bridges making the gameplay extra fun as you don't know what's awaiting for you in the next building.

screenshot


SoulS33ker:
rg
hint = hint ?

LuismiInTheZone:
The ores are ok, but I have one thing that I don't understand very well, why does the blue team have the 3rd mese block at the bottom of a building while team red has the mese block at the top. I would suggest putting all the mese blocks at the same level, so like that some teams doesn't have advantage of getting there faster.

Red: 
red

Blue: 
blue



Suggestion:

1. Hide a bit the mese ores, for me at least seeing a mese block on the top of a building in the middle of no where seems a bit ugly, maybe you can find some place (easy, not very hidden but not left in the outside) where can be found. You could also include, if you do this, something in the hint. 

Example: you can put them inside of this.
example


2. Add iron and coal. It is recommended since if in some case the team is quite bad in amount of iron swords and they lost all mese would be good to have something left. One suggestion here is to place them around destroyed buildings.

SoulS33ker:
> I would suggest putting all the mese blocks at the same level

replying to luis
its probably fine,appleton has mese blocks for 1 team on theattic and other one hidden in a safe room underground  ¯\_(ツ)_/¯

LuismiInTheZone:
> > I would suggest putting all the mese blocks at the same level
> 
> replying to luis its probably fine,appleton has mese blocks for 1 team on the attic and other one hidden in a safe room underground ¯_(ツ)_/¯

Appleton has like 50 mese blocks, this has only 3 per team

HobbitPower:
@ChandlerBing1994 I agree with Luis. Apart from the ore placement, I have nothing to criticize. Put the meseblock at the same position as on the other team side. About the iron: Luis has a good idea. You can hide 4 iron/coal per structure here. 
image
image
image
That way they are not visible and don't destroy the cityscape. In my opinion, the mese blocks can remain visible. Perhaps it would be advisable to add a hint about the iron ore if you want to hide it.

Fix that and I will approve it :)

SoulS33ker:
no 🐛
all good :thumbsup:

LuismiInTheZone:
It seems to be okay. The only problem I see is:

Blue team: 26 iron, 48 coal, 17 mese
Red team: 33 iron, 55 coal, 17 mese

The coal and iron counts are a little different on each team.

You can try putting a little more ores around where is the rest and that's it.

HobbitPower:
@ChandlerBing1994  Apart from the small difference in the iron ores, there do not appear to be any other discrepancies. Just adjust it and I'm happy ^^

MoreKat (commented):
The map looks awesome and is unlike any map that we already have. My only remark is to add blocks or ladders to initial_stuff to make it easier to climb up the buildings and navigate.
Not sure why it is not enabled for Nade Fight, I think it would be fun, but that is up to you anyway. If the matches in this map end up being too long it might be good to increase the speed, since players have to cross quite some distance, even vertically, but I think this is something that we have to check when the map has been played.

Themostrandom:
Hi, 

I really like your map, I have 2 suggestions and no comments: Add coloured wool to the team building (not essential) and add ores (not essential either).

ClothierEdward:
I like the idea of this map. It makes itself distinct from the other city maps through having really tall buildings and having a post-apocalyptic aesthetic. I like the barricades in the map that contain ores, it's a really creative way to obtain ores. I also like the fact that there's a few bridges on the map to make traversing it easier. There are a few issues that need fixing though:
1. The map needs more destroyed buildings. I find it weird that there's only one destroyed building on the entire map despite it being named "Abandoned City". The map could do with a couple of destroyed buildings which could add to the challenge of the map while improving its aesthetic.
2. Make the map available to play on Nade Fight. I find it so weird that Nade Fight is disabled for map, despite it being the perfect type of map for that mode. Just imagine being able to scale buildings with the knockback nade while battling players inside the buildings. That would be lots of fun.
Other than that, I think this would become an instant classic due to its challenging nature. I would love to see this map come to CTF in the future.

PR #339, opened 2024-06-04
Add new map "Coral Street Bridge"

AnonomousAccount:
Ummmm ignore all the commits except for the last one.
It's basically just all from my other PR, and then removing those files
Edit: This map is currently WIP, I have a buncha gameplay problems to fix, suggestions apreciated

PR #355, opened 2024-07-25
Add new map "Coral Street Bridge" by Anonomous

AnonomousAccount:
**This is the second PR. The last one was closed due to GitHub shenanigans**

Issues posed by map testers on last PR:

- Map is too thin


AnonomousAccount:
This map is kinda similar to the newly-merged Golden Gate Bridge, though I believe the experience will be pretty different. PVP will be more centralized and speedrunning should be less common.

HobbitPower:
Continuation from #339

HobbitPower:
Tested it with the latest CTF versions (Github and CDB): https://discord.com/channels/447819017391046687/448136831385468930/1268538815971197000
Maybe more will test the map and share their experience with this.

Nicely built otherwise, but honestly very similar concept to the Golden Gate Map.

LuismiInTheZone:
You forgot to add the license. Did you change the .conf file by yourself? Because the license part doesn't appear

LuismiInTheZone:
I have the same problem as HP, I don't know why. My opinion is that the map is fine, but I would light up the part under the bridge. Another thing is that in some vehicles some of the blocks are indestructible and others destructible. I recommend choosing one or other. Also, if you want, it is advisable to expand the chests around the map as well. The map is very similar to the Golden Gate Bridge, so it could have problems in the future.

LuismiInTheZone:
Try to export the map again without changing anything (only the necessaries and things to change) and put the map again

LuismiInTheZone:
> I have the same problem as HP, I don't know why. My opinion is that the map is fine, but I would light up the part under the bridge. Another thing is that in some vehicles some of the blocks are indestructible and others destructible. I recommend choosing one or other. Also, if you want, it is advisable to expand the chests around the map as well. The map is very similar to the Golden Gate Bridge, so it could have problems in the future.

Also, please remove the chest you put by yourself in the small boat, it can cause problems

AnonomousAccount:
Unfortunatly, this map was too similar to another map

PR #356, opened 2024-07-25
Add new map "Two Pagodas" by Anonomous

AnonomousAccount:


HobbitPower:
Continuation from #353

SoulS33ker:
no bugs,equal ores
flags,barriers,chests all good
interesting concept to give 4 openings to take flag other than the main entrance

Themostrandom:
I like your map but I have several brands: 
only one entry in the base
some blocks can be changed (tree trunks by real trunks)
the size of the map, which in my opinion is too small
I suggest removing the wooden panels, as players can write on them.



I like the style of your map but I think it's too small for ctf if there are 25-30 players.

AnonomousAccount:
Ah I didn't see those grabbing points at the top. I will remove those and rework the bases.

AnonomousAccount:
Currently putting this project on hold: can someone pls add WIP tag?

AnonomousAccount:
Update:
-Changed base designs to hopefully allow for more defence

Themostrandom:
Hello, 

I like your map, I think it's speedrunable, but it's not serious. Overall, no problems, but I have a question about the blocks on the ground floor, some are breakable, others are not, and the upper floors are unbreakable.

AnonomousAccount:
Yes, that is by design. Players are able to break into all the sides of the bottom to allow more entrance to the pagoda

AnonomousAccount:
WIP can be removed now, map is ready for reviews now

AnonomousAccount:
Base Update is planned, just procrastinating on it rn

AnonomousAccount:
MAP MANAGERS: Would you prefer I keep the current cherry tree design (jungle wood trunks and pink wool leaves), or would you be fine with custom cherry tree blocks?

Themostrandom:
> MAP MANAGERS: Would you prefer I keep the current cherry tree design (jungle wood trunks and pink wool leaves), or would you be fine with custom cherry tree blocks?

Unfortunately, I don't think any blocks can be added :/.

AnonomousAccount:
You can add custom block with an init.lua

ClothierEdward:
This map is pretty neat. It's a small map with indestructible pagodas to make speedrunning the map harder. Really the only change I'd make to it is to add something on the top floor. I find it weird that there's nothing there. Other than that, a pretty good map.

gameit0:
The screenshot is not placed in the map's folder.

AnonomousAccount:
Yes, thank you, fixed now

AnonomousAccount:
Map is on JMA.
From playing it in real matches a few times:
- Its pretty fun with 10-25 players. Any more and it starts to get a bit crowded
- Smaller player counts aren't as detrimental as other maps because its easier to find players
- Bases offer a great balance of defendability, without being too easy or too hard to break in

PR #344, opened 2024-06-26
Snowland Chest Spawn Fix

AnonomousAccount:
This is a simple change to the map.conf of "Snowland"

• Chests no longer to spawn outside of the border, and out-of-reach of players
• One chest zone covering the whole map has been divided into two, giving each team an equal number of chests every time
• Chests can no longer spawn in underground dungeons, which were out of sight from players

This has no effect on the number of chests which spawn

PR #357, opened 2024-07-25
Fix Chest Spawn on Snowland

AnonomousAccount:
This PR fixes a few issues with the way chests spawn on "Snowland"

- One chest zone that used to cover the whole map has been divided into two, giving each team an equal number of chests. The total number of chests has not been changed.
- The chest zone has been shrunk so chests no longer spawn outside of the border, and therefore out of reach of players.
- Chests no longer spawn in underground dungeons


HobbitPower:
Continuation from #344

PR #359, opened 2024-08-22
Add the 'Abandoned Mine' map

kellymbriggs:
i have added a new map titled 'Abandoned Mine'. IGN: slimy_bannana_peel


Themostrandom:
I have a lot of comments, starting with the size of your map, which is much too small (41*49). The same is not very rp in the mine style (meselamp) There are carts that lead to nothing and for a mine the fact that you can not break any block is disturbing more there are torches on mese lamp, holes in the walls with glass ... etc..

I like the idea of a map like this one, but it was made far too quickly and is therefore unplayable and non-rp.

kellymbriggs:
Allright, this map was my first map, so it's not perfect. I wanted to make a smaller map, because almost all maps on the server are quite large.

1. There are no unbreakable torches, and I wanted the map to always be lit.
2. The torches placed on mese lamps are to differentiate red/blue. Red has just mese lamps, while blue has torches placed on mese lamps. (there are obviously better ways to differentiate red/blue)
3. "carts that lead to nothing". what are carts?
4. "holes in the wall with glass". those were supposed to be exits to the outside, but I sealed them off because I don't want an outside.
5. "and for a mine the fact that you can not break any block" - i didn't want this map to be like tunnel/caverns where tunneling is encouraged.

I will delete this map and make a better/bigger one.

Thanks for the feedback!

kellymbriggs:
Allright, this map was my first map, so it's not perfect. I wanted to make a smaller map, because almost all maps on the server are quite large.

1. There are no unbreakable torches, and I wanted the map to always be lit.
2. The torches placed on mese lamps are to differentiate red/blue. Red has just mese lamps, while blue has torches placed on mese lamps. (there are obviously better ways to differentiate red/blue)
3. "carts that lead to nothing". what are carts?
4. "holes in the wall with glass". those were supposed to be exits to the outside, but I sealed them off because I don't want an outside.
5. "and for a mine the fact that you can not break any block" - i didn't want this map to be like tunnel/caverns where tunneling is encouraged.

I will delete this map and make a better/bigger one.

Thanks for the feedback!

kellymbriggs:
I don't have discord (this also isn't my GitHub account), however, you can message me in-game.

HobbitPower:
@kellymbriggs For the future: You can also leave the PR open and edit/update the files after there is feedback. In this case, it's probably fine since you want to submit a completely new and bigger map. But if there are only minor things to change, you can replace old files of your PR with newer ones without having to close the PR.

AnonomousAccount:
Hi silmy, I wanted to help you with this map, if you are ok with that.

How can I contact you, do you have discord, matrix or some other rapid-contact method?

AnonomousAccount:
@kellymbriggs

PR #361, opened 2024-08-26
Add new map "The desert war" by Themostrandom01

Themostrandom:
This is a desert-style map with sniper posts, trenches and minerals stolen from the enemy. The screenshot is in the right format, and the speed is slightly higher than on my other maps :).


SoulS33ker:
image
i dont know what else to say ¯|_(ツ)_|¯

Themostrandom:
> ore imbalance iron 9:16 mese 3:7 diamond 9:15 screenshot is not 3:2

ore are okay : they can have the same acces if i'm good
*

SoulS33ker:
> > ore imbalance iron 9:16 mese 3:7 diamond 9:15 screenshot is not 3:2
> 
> ore are okay : they can have the same acces if i'm good *

It's about being available during build time,Doesn't matter if equally available to both team once barriers are removed.
unless i am mistaken about what you meant to say

SoulS33ker:
oh oj


> > ore imbalance iron 9:16 mese 3:7 diamond 9:15 screenshot is not 3:2
> 
> ore are okay : they can have the same acces if i'm good *

oh my bad,all good then :)

SoulS33ker:
> > ore imbalance iron 9:16 mese 3:7 diamond 9:15 screenshot is not 3:2
> 
> ore are okay : they can have the same acces if i'm good *
destructible block

i can see what you were trying to do,but you placed destructible acacia planks,i can get in in build time.

MoreKat:
I haven't tested this map yet but the screenshot does not give a good overview of the map. You should take it from a higher angle.

Themostrandom:
> I haven't tested this map yet but the screenshot does not give a good overview of the map. You should take it from a higher angle.

it show all of map and i think you can see all

SoulS33ker:
ores are balanced now,no bugs seen,good map
,not sure about the screenshot decimal values seems to be involved,someone else can check it

Themostrandom:
> ores are balanced now,no bugs seen,good map ,not sure about the screenshot decimal values seems to be off,someone else can check it

its 3:2

HobbitPower:
> it show all of map and i think you can see all
* You could definitely take better screenshots though. Also yes, the ratio is a little off. 1526x1017 = 1.50049... (instead of 1.5). So when rounded, the size fits more or less, although a better screenshot in general wouldn't hurt imho.

* Now to the bigger point of criticism: One team can still reach the ores more easily than the other. I see the blue team at a slight disadvantage there. I would also edit the hint (e.g. There is a chamber with ores in the centre or something similar).

lazylier:
I think this map is going to be very imbalanced in classic. One team will dominate the mine and the other will be left with nothing. The Wall is already like this, but it has two separate mine entrances so it's not so bad. Otherwise it looks like a good map. I just don't know about the mine as it is right now.

ClothierEdward:
A new desert map hmmm? I think this map makes itself distinct with the labyrinth in the center of the map. Other than that, the map is pretty decent. I didn't find any issues with the map while playing it. The map works and the map has good ore distribuiton. A pretty fine map, I gotta say.

PR #362, opened 2024-08-27
New Map: Chaotica

mrtechtroid:
New Map: Chaotica
- For Nade Fight, and Classes
- Uses Team Yellow and Purple. (CAUTION: Please use the latest fork of CTF, and not from ContentDB since Team Yellow is not updated there)
- Low Gravity - To Jump Platforms. 
- Small Map. 

Would love your feedback on this map. 

screenshot


Themostrandom:
image
Several problems celon moi: Il y a 2 teams chest cela peut causer des bugs et n'est pas utile(mais surtout cela double la quantité de stuff stocké et cela correspond a votre map) The size of the map is 115*55, and there are not enough chests for a map I call chaotic. I suggest adding a block at the player's spawn.
I like the idea of these platforms, but I'd have put more over a greater distance (x,y,z axis). --> It comes back to the problem of map size. There are no minerals, but that's not a real problem because there's no classic mode (in my opinion, 5 diamonds on each side would be good for class axes).

I like the idea of your map but the main problem for me is the size of the map and the 2 chest teams.

chaotica/map.conf
Themostrandom, line 1: game_modes = return {"nade_fight","classes"}
barrier_area = return {pos1={x=0,y=0,z=0},pos2={x=115,y=87,z=55}}
teams = local _={};_[1]="flag_pos";_[2]="enabled";return {yellow={[_[2]]=true,[_[1]]={x=111,y=19.5,z=27},pos1={x=59,y=0,z=1},pos2={x=115,y=87,z=55}},purple={[_[2]]=true,[_[1]]={x=6,y=19.5,z=26},pos1={x=0,y=87,z=55},pos2={x=58,y=0,z=1}}}
chests = return {{amount=10,pos1={x=108,y=14.5,z=15},pos2={x=112,y=17.5,z=39}},{amount=10,pos1={x=5,y=14.5,z=15},pos2={x=9,y=17.5,z=39}},{amount=10,pos1={x=54,y=14.5,z=22},pos2={x=58,y=20.5,z=31}},{amount=10,pos1={x=60,y=14.5,z=22},pos2={x=63,y=21.5,z=31}}}
phys_jump = 1.5
phys_speed = 1.1
size = return {x=115,y=87,z=55}
map_version = 3
enabled = true
name = Chaotica
phys_gravity = 0.30
initial_stuff = return {"default:ladder 50", "default:cobble 99"}
author = superbull
hint = Hope for no lag.
license = MIT
others = 
treasures = 
skybox = none
enable_shadows = 0.26
start_time = 18600
time_speed = 1

mrtechtroid:
Hi there, @Themostrandom 
The Two Chests are "Intentional" unless they break something in game. 
Its supposed to be a small map, for lesser no of players. Similar to the house map. 
I will be adding more chests to the map to make it better. 

Regarding the distance between platforms, it is kept lower to allow players to actually be able to jump on the other platform without missing.

SoulS33ker:
must change:
1.the team zones are select wrong,should be an easy fix to re-select them
https://github.com/user-attachments/assets/bc6d1558-d492-401b-a636-3affd1cc4c8f
others opinions:
1.we usually do 5x5 protected area around the map,if map managers agree with the small area of your spawns then its fine
2.any reason for the thin map?
3.i see no problems with the extra team storage space  👍
4.The pvp-junkies in ctf would be mad i think,as there's less space of pvp,towering would be useless,might fall to death,but personally i would love to see this map and new type of pvp  being born here.

SoulS33ker:
i would also like to how did u choose the locations for the platforms,are they calculated strategic locations or so was it just random?

mrtechtroid:
@SoulS33ker 
- I'll fix the team zones later. 
- If there are objections with 5x5 area, ill try fixing. (once again lets put some chaos)
- Same size(sligthly larger in width) than Sams Mansion
- Platforms were randomly placed, with additionally making sure that every platform can be accessed by at least one platform without boost, and at least 2 with boost.

SoulS33ker:
@mrtechtroid 
could it be difficult to capture flag due to :
1.narrowness of map,not providing a meta to sneak into the base(i dont belive in teamwork)
2.extra base making materials in 2nd team chest

we dont have a thin map to back up my points thats why i want one of these thin maps in,sams mansion provides tunnels and enough cover for players to escape the eye & sneak in (it is to be noted it is a speed run map,not really that well liked)

mrtechtroid:
@Themostrandom @SoulS33ker @HobbitPower
- Fixed team zones. 
- Added ores on the map now. (Wanted to ask if its enough to be allowed to enable classic now)

AnonomousAccount:
**Gameplay Comments:**

Remember, players won't go places that have no benefit to their gameplay. Such will happen with the higher platforms. I also see this map as being frustrating (falling to your death and losing all your items tends to be). 

The gameplay I would expect would be each team bridging in a perfectly straight line towards the middle and the whole match being bridge PVP. While some enjoy this, there is no alternate option. With 20 or more players online, this map would most likely last hours, the items dwindling as they fall into the void.

The concept _might_ work, but this current approach will not

AnonomousAccount:
Also, low gravity doesn't work that well in nades. Nades need to collide with an object to be effective, and if players are flying through the air, they will be easily not be able to be hit. 

Bassicaly: Nades + Low Gravity = Melee fight

mrtechtroid:
@AnonomousAccount you do NOT die instantly on hitting the floor, if you jump instantly as soon you fall down, you will reach one of the lower platforms, with the help of ladders. 
Players will go to higher platforms if they intend on sneaking and getting the flag. 
Nades have an radius where they will give damage to players, and there is always void bombs to bring you down. 
This map is again meant for smaller player sizes.

AnonomousAccount:
It's maps like these in which I wish we had a testing period

HobbitPower:
@mrtechtroid 

* Please fix the team zones. I get teleported back when I'm one block away from the red barrier. 

* I would recommend removing the second team chest.

* I like the idea of changing gravity as it ties the platforms in better. But I would disable the map for nade fight then. It's not enabled for classic btw?

* The hint is completely unnecessary. You could use it instead to point out the ores.

* Speaking of ores: I would rather make the ceiling destructible instead of the wool (floor and outer walls).

* Small things that don't necessarily need to be adjusted, but can be: When the red barrier disappears there is a gap in the kill nodes. Also a block in the outer barrier is missing.

* It's an interesting concept, very experimental. It is very possible that many people will not like the map though. I would agree to a test phase if you are willing to make changes, given enough feedback.

mrtechtroid:
Changes performed: 
- Centered the flag now and added more symmetry. 
- Made the center 2x4 blocks in the middle/central section breakable. 
- Fixed the barriers, outer walls, and added the upper roof. 
- Added all three modes to be allowed for the map. 

No changes on:
- No of team chests per team is still 2. 
- Size of the map is still the same.

HobbitPower:
The biggest bugs have been fixed, so the action needed label has been removed for the time being. 
Feedback still welcome, so no final approval yet.
As already said: Interesting concept, although I could imagine that many might not like it (platforms, gravity, size of the map). Personally, I also think that the map is quite small and that speedruns are possible, at least in nade-fight. 

So I rather see an approval for a test phase on the main server. If there is too much negative feedback, changes will have to be made.

Themostrandom:
Bonjour, 

Your map still confuses me with these 2 team chests :). It seems fine to me, I just have 2 questions: why did you choose these team colours and why did you decide to put the night from the start.

mrtechtroid:
Wanted this to be the first map using unconvential colors (rather than red and blue). The lighting with the lamps looks nice during the night. So nighttime

ClothierEdward:
Okay, the idea for this map is really creative. I really like the use of floating islands and the increased gravity is helpful for making it across the map. There's only a couple of issues I have with the map.
1. Having two team chests are as helpful as they are confusing. I'd suggest adding signs on top of the team chests to label each chest as "Team Chest A" and "Team Chest B".
2. I find it weird that you can only dig through the top center island through the center portion of it. Make the sandstone bricks mineable so that we can reach the ores in the center island more easily.
3. The amount of ores on each side are equally underwhelming. 8 iron, 7 mese, and 5 diamonds are not something you can really work with on classic. I'd suggest adding more ores to the center island so that playing on classic can be more viable.
Make those changes and I think this could be a pretty good map.

PR #365, opened 2024-10-10
Add "Pumpkin Hills" map by ClothierEdward

ClothierEdward:
This map is a Halloween-themed map that would be exclusive to the Halloween event that happens every October. It is also intended to be used with the Halloween texture pack that is enabled during the Halloween event.


mpixelate:
The pumpkins look a bit robotic. I think you should add more details, especially a mouth. The map is also quite dark. I recommend searching for pumpkin references on the internet for inspiration. Below is my version of the pumpkins.
screenshot_20241010_110411
screenshot_20241010_110416
screenshot_20241010_110420

winniepee:
After playing for quite some time, here are my thoughts:

The map looks very nice. It does fit in with the Halloween theme. Haven't stumble on an error, so that's an OK. The ores in each of every pumpkins are perfectly equaled, the only problem is ... no coal? I found it weird to not have coal in some of the pumpkins. Yes, you can still use wood, but I think having coal is a must. 

Both base are balanced, but I do have something to suggest. Maybe add some windows for blue's base cuz red's base has more than blue's.
Here's the comparison:
screenshot_20241011_000028
screenshot_20241011_000050
Also make both houses breakable, giving players easy access into the houses.

I do hope that my suggestions can help a bit. Wish it will be published when the event happens 😄.

AnonomousAccount:
Map Design Review:

Layout: Pretty simple, yet effective layout. You _could_ add a structure in the middle of the map for more clashing.

Base Design: Great houses. One thing though, is that there are only a few ways to enter the flag room. You need less than 50 blocks to seal off the flag room.

Loot Distribution: Chests are good, pumpkin ores are an awesome idea

Gameplay: The map might be a little hard to navigate, given how dark it is. I would suggest lighting up the middle of the map some more (the pumpkin plantation)

HobbitPower:
@ClothierEdward The pumpkins can stay like this for my sake. The theme is simple, but fulfils its purpose. 

Please note the suggestions already mentioned (windows/entrances, coal etc.). I know the map is supposed to be dark, but maybe you can place some fires here and there? I think that would still fit the theme and would help to navigate.
Screenshot should also show the whole map including both bases.

ClothierEdward:
Just updated the map after looking through the current feedback. The pumpkins now have various expressions on them along with blocks of coal on the inside. On top of that, there are lamp posts dotted around the map to make it easier to see. As for the blue team's base, the base has more windows now to make it easier to break in. Those are all the changes I have made to the map thus far.

ClothierEdward:
> Base Design: Great houses. One thing though, is that there are only a few ways to enter the flag room. You need less than 50 blocks to seal off the flag room. Personnaly, I really like this, but I just wanted to point this out because there are some who do not. Edit: Just noticed the bases are not balanced, as par Winnie's suggestion, you really need to add more entrances to blue's


I just updated the map so that the blue base has more windows which should make it easier to break in. The idea with this map is that you're supposed to break in through the windows of the mansion which is why you can't mine out the rest of the mansion. Also, for any eagle-eyed viewers, you might have noticed that each base has a secret entrance. Where might they be? Only one way to find out.

lazylier:
I have two complaints. First, there's no roof to the map. That needs to be fixed. Second, red has a little more iron than blue (64 and 56). That might not matter, but it's there. Otherwise it looks great!

LuismiInTheZone:
Ok, here is my review:

The bases are really amazing, I love the details you gave to those mansions. In my opinion I really wish the big pumpkins could be a bit different, but since this is a square game/there are not many blocks compared to minecraft I think it's good enough and doesn't need a change. 

The only mistake I see is the one that binshaden said, about iron that one team has more than others.

My mayor suggestion for the map is to add more decorations/small houses/big houses around the sides of the map, specially on the red side, because for some reason blue part has an amazing tower but the red part has nothing (I think, I haven't seen anything). Also, one of the sides, compared with the other, has way more trees than the other, and in my opinion, would be nice to add trees to that side without those much trees as the other.

Other suggestion that I only recommend you doing if you have a lot of free time (since it might take a while):
- In the caves that are going underground, maybe you could do something interesting like was made in the map Appleton, but as I said, this might take a lot of time to think about what to do and how to do
- You can try testing more/less light in bases, but that needs coding and I haven't tested this before, so if you don't want to try new things just ignore this.
- This is very hard too, but you can try creating a skybox with more detail than the current one, would be nice to see another skybox with some deep dark sky, but at the same time that has some light (I hope you know what I mean, if not here you have a picture: )
0120jawr

HobbitPower:
> Ok, here is my review:
> 
> The bases are really amazing, I love the details you gave to those mansions. In my opinion I really wish the big pumpkins could be a bit different, but since this is a square game/there are not many blocks compared to minecraft I think it's good enough and doesn't need a change.
> 
> The only mistake I see is the one that binshaden said, about iron that one team has more than others.
> 
> My mayor suggestion for the map is to add more decorations/small houses/big houses around the sides of the map, specially on the red side, because for some reason blue part has an amazing tower but the red part has nothing (I think, I haven't seen anything). Also, one of the sides, compared with the other, has way more trees than the other, and in my opinion, would be nice to add trees to that side without those much trees as the other.
> 
> Other suggestion that I only recommend you doing if you have a lot of free time (since it might take a while):
> 
> * In the caves that are going underground, maybe you could do something interesting like was made in the map Appleton, but as I said, this might take a lot of time to think about what to do and how to do
> * You can try testing more/less light in bases, but that needs coding and I haven't tested this before, so if you don't want to try new things just ignore this.
> * This is very hard too, but you can try creating a skybox with more detail than the current one, would be nice to see another skybox with some deep dark sky, but at the same time that has some light 

Yes, I think these are ideas for the future, perhaps as an update for next year. The most important thing right now is to have the same number of ores. Sure he can add decorations, trees etc. but only if has the time. Btw you can hide light sources under slabs/stairs and it will still light the area around it, very useful. And a skybox is really not that easy to make, so that might be an option for the future, but I also think that the current skybox is sufficient imo.

HobbitPower:
I think the map is alright now, also because it was created at such short notice. As already mentioned, changes for next year can be made after the event. More details, buildings etc. I think you could also cut parts of the underground if it is not used for anything in the future. 
Ores are fairly distributed, chests too, similar number of entrances to the base (we'll see how fair the two different mansion types are). The tower is a small advantage for team blue (e.g. for snipers), but I'll ignore that this time. Can and should be adjusted for the future. You could also upload a better screenshot, but that is of secondary importance now.

LoneWolfHT:
Merged into the seasonal_halloween mod https://github.com/MT-CTF/seasonal_halloween/commit/4282df62f95b704036349575ed2895d03cde6d82

PR #367, opened 2024-10-28
Balance ores in Caverns

SoulS33ker:
**1.** The current Caverns has an unequal amount of mese blocks; red has 7 & blue has 0.Of the 7 blocks on the red side,3 can be found on way to middle of the map,the location of other 4 is not very obvious.
This PR add 7 blocks to blue side and relocates the group of 4 blocks of red to a more obvious location.
**2.** The actual mese and diamond ores,the ratio currently stands at diamonds 72:39 & mese 101:54 for red & blue respectively.
The PR changes the ore ratio to diamonds 50:50 & mese 50:50,much of the old locations are retained ,while those which were random and didnt make sense were removed while bringing down the number,the locations of the new ores are expected to be " obvious ",it can be found as rewards for digging tunnels & clearing lava,surface ores to hint at veins below and at air pockets (these already existed and is expected to be easily found through  "camera gitch ".
in conclusion:
**1.**  Red team;3 mese blocks on way to middle of the map(unchanged),2+2 in dungeon rooms(new,re-location).
**2.** Blue team;3 mese blocks on way to middle to the map(new),2+2 in dungeon rooms(new).
**3.** 50 mese ores and 50 diamond ores per team


SoulS33ker:
for ease of testing,you can convert the the mese ore etc into  "mese lamp",will be easier to find ores while you noclip.

Themostrandom:
Hello, this modification to the map seems correct.

PR #368, opened 2024-11-03
Make blocks found in chests more appropriate for certain maps

AnonomousAccount:
Specifically on maps that spawn players with blocks different than "chest blocks" (cobble or wood), matches chest blocks to the default block.

Maps currently affected:
- Ewok Village (All wood in chests is now jungle wood)
- Nether Kingdom (All cobble in chests is now desert cobble)
- Pyramids of Akhenaten (All cobble in chests is now desert cobble)

### DIFF
--- a/ahkmenrah_pyramids/map.conf
+++ b/ahkmenrah_pyramids/map.conf
@@ -8,7 +8,7 @@ time_speed = 1
 size = return {z=218,x=218,y=60}
 hint = There is an invisible labyrinth under your base and a treasure at the end of it!
 others = 
-treasures = return {"none"}
+treasures = default:cobble;0;0;0;0.0;1;default:desert_cobble;50;99;2;0.3;1;
 skybox = none
 start_time = 5900
 phys_speed = 1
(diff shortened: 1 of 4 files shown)


AnonomousAccount:
Would appreciate feedback, and also if any other maps should be included/excluded

AnonomousAccount:
What are opinions on the change on Nether Kindgom? I am conflicted about it, because one upside of cobble is that you can tell what blocks have been placed by players and which blocks are naturally there.

lazylier:
> What are opinions on the change on Nether Kindgom? I am conflicted about it, because one upside of cobble is that you can tell what blocks have been placed by players and which blocks are naturally there.

My opinion is it doesn't make much difference because smart players will already have plenty of matching blocks to use when it matters.

AnonomousAccount:
Why have to carry two types of wood when you can only carry one, for example. In the same way that all trees craft into apple wood, just so you don't have a buncha different types of blocks in your inventory

PR #369, opened 2024-11-30
Fix "Terrarium"

winniepee:
There's a hole near red's border (picture for reference), which I have fixed. I also change the map's corner just in case.

screenshot_20241130_143534

Video: https://discord.com/channels/447819017391046687/533289338554023939/1312297189736972369


HobbitPower:
@winniepee You get these errors as well? Re-saving could solve it.
image

image

image

AnonomousAccount:
Don't you just need to change the mts? not the conf or barrier file

HobbitPower:
Seems fine now.

PR #371, opened 2024-12-05
Add new map: Serpent's Pass Reimagined

ScaleRaker:


Themostrandom:
Hello, the boundaries of team territories are incomplete (we've been moved back to the spawn 2-3 blocks away). Your map looks like the dragon dance map :) . I like it but I have a suggestion: add chests closer to the spawn.

ScaleRaker:
What do you mean by "we've been moved back to the spawn 2-3 blocks away"?. As for the chests, I'd like the players to not just make use of the main bridge but also others. If they collect loot somewhere further away then they will be less likely to keep making use of the main bridge and split up.

Themostrandom:
> What do you mean by "we've been moved back to the spawn 2-3 blocks away"?. As for the chests, I'd like the players to not just make use of the main bridge but also others. If they collect loot somewhere further away then they will be less likely to keep making use of the main bridge and split up.

Team zones are poorly defined

ScaleRaker:
You are right, I'll fix that.

gameit0:
From what I have seen, it's a solid map without issues. :+1:

ClothierEdward:
This map looks pretty cool. It reminds me of maps like Dance of the Dragons and Nether Kingdom. I also like the use of reinforced cobble, a block you rarely see in CTF maps.
I do think this map is missing something though. There should be some towers in the center, perhaps with some ores since this map is available in classic. Alternatively, adding ores to the back walls of the map would be quite nice.
One more thing, making this map unavailable in nade fight is an interesting choice. Whether or not you want to enable the map in nades is up to you.

ScaleRaker:
Thanks for reviewing the map. 
First of all, the inspiration. This was originally supposed to be a Serpent's Pass-type map with emphasis on removing the not-so-fun parts. Now it's quite different with a relatively good amount of inspiration from a LOTR build I did some years ago.
As for the ores, they are already placed at the bottom of the stairs near the chests.
This map could potentially benefit from some towers near the middle.
I dislike the way Dance of the Dragons on nade fight feels, or really, any bridge nade fight pvp. Therefore, I turned it off.

magma_strongholds/map.conf
```diff
@@ -0,0 +1,21 @@
...
+phys_speed = 1
+size = return {z=255,x=150,y=72}
+map_version = 3
+enabled = true
+name = Magma Strongholds
+phys_gravity = 1
+initial_stuff = return {"default:pick_mese"}
+author = ScaleRaker
+hint = Treasure is hidden under the lava!
+license = CC-BY 4.0
```
LuismiInTheZone: I don't see any mistakes, just the license, needs to be CC BY-SA 4.0. The map is amazing, I hope it gets added soon

magma_strongholds/screenshot.png
LuismiInTheZone: Ah yeah, also another thing, the screenshot needs to be 3:2, try converting it with some web page, it should be easy

PR #372, opened 2024-12-09
Add "Snow Globe" map

ClothierEdward:
This map is a Christmas-themed map that would be exclusive to the Christmas event that happens every December. The map is a giant snow globe featuring igloos, a pine forest, a festive workshop, and a christmas village.

(Now in a seperate branch)


AnonomousAccount:
Would be really cool if there were ambient snow particles

Themostrandom:
Hello, this map looks good since I helped you make it ;) . However, I have a comment and a suggestion: the chest zones are very tight and there aren't many chests (30 per size). I think you should add a few more and place them all over the map.

LoneWolfHT:
Merged into xmas repo (Will be added for the full month of December, extra during the winter event) https://github.com/MT-CTF/seasonal_xmas/commit/59a542c272ccffd30f8bfb20483bbc5eebaf41fc

From a quick check it seems like the map is a little bit on the smaller side, will have to see how it goes.

PR #373, opened 2024-12-11
Add the world trade center map

Quietq2048:


AnonomousAccount:
Could not load map

AnonomousAccount:
been stuck on the loading screen for five minutes ._.

HobbitPower:
> been stuck on the loading screen for five minutes ._.

Seems to be a you problem, I can load the map. 

* But I get this error message. Normally this can be fixed by re-saving the map. 
image
* Team Blue seems to have a slight advantage because the base is inside and not outside a building.
* I'm actually afraid that the towers as such aren't really used much, so it's good that there are mese and diamonds hidden in one of the towers. I've only found the ores in the antenna so far, does the other tower also have an ore spot? If so, where?

* (Are the meselamps in the Obsidian buildings there for lighting or also for design reasons? If the former, you can hide light sources under slabs/stairs etc and the light shines through there as opposed to whole blocks. Just a tip.)

Just a first short test, I'll take a closer look at the map again.

Quietq2048:
> > been stuck on the loading screen for five minutes ._.
> 
> Seems to be a you problem, I can load the map.
> 
> * But I get this error message. Normally this can be fixed by re-saving the map.
>   image
> * Team Blue seems to have a slight advantage because the base is inside and not outside a building.
> * I'm actually afraid that the towers as such aren't really used much, so it's good that there are mese and diamonds hidden in one of the towers. I've only found the ores in the antenna so far, does the other tower also have an ore spot? If so, where?
> * It is recommended to have at least one chest zone per team. This ensures that both teams have exactly the same number of chests. (So 2 x 32 in your case)
> * (Are the meselamps in the Obsidian buildings there for lighting or also for design reasons? If the former, you can hide light sources under slabs/stairs etc and the light shines through there as opposed to whole blocks. Just a tip.)
> 
> Just a first short test, I'll take a closer look at the map again.

if you dont mind, can you please reupload the map, and perhaps put ur name in the map too? i have alot goin on in life and i do not have any more time.

Themostrandom:
image
image
image
image

Hello, several comments: the spawn chests in the trees. the inside of the black buildings are very dark, I don't know if this is really the case but the fountain is not centred ... There are some strange holes in your buildings and map barrier. On such a large map, I think the addition of more chests would be positive.

PR #374, opened 2024-12-12
Add Four Corners map

lazylier:
This is a four team tunneling map. There is a lot of iron around the middle of the map, as well as a huge hidden chest zone and four hidden treasure areas. If necessary, I might add more iron near the base exits.


AnonomousAccount:
Screenshot 2024-12-28 at 7 10 13 PM
Not a priority, but the OCD nut in me hates this wrongly oriented brick lol
Did you have redef on while taking this screenshot?

lazylier:
no, I did not

Themostrandom:
image
image
Hello, your map looks fine to me, the red barriers are missing. and the ores are in odd numbers (63) so there will be one ore left.

ScaleRaker:
I don't personally see a problem with an odd number of ores. I really don't think anyone is going to mine every single ore out, but, I mean, that's just my opinion.

lazylier:
There should still be an even number of ores overall since there are four treasure areas. I've added barriers to the air sections. Does it work now?

gameit0:
Looks like main to play. There are no connections between the chest-structures and/or the bases. Players need to dig for ages before they will meet another base or something else. Another point are the bases themselves, they only can be reached by 3 pre-defined spots as the rest of the walls are our of indestructible blocks. The 3rd problem I see are the huge vertical distances: Digging usable vertical tunnels/giant stairs will take even more time than "normal" tunnels.

lazylier:
Thanks for looking at it! I probably should shrink it more. It's based on the existing Tunnel map, so I doubt the bases would be any more of a problem than they are in that one.

The 3rd point is the most difficult. Right now it's a cube, so each flag is an equal distance from each of the others. Vertical digging is longer, so maybe it should be squashed a bit, but to do that, I'd want to make the effective distance between flags the same, if possible.

PR #375, opened 2025-01-09
Add map "Battle in the North" by slimy_banana_peel

AnonomousAccount:
Map by slimy_banana_peel.
Contact them using in-game messaging


AnonomousAccount:
Map needs a newer screenshot

AnonomousAccount:
Map is very large. Would benifit nicely from a "min player count" rule

lazylier:
I wish the team areas were the same size. Green's is a lot bigger

AnonomousAccount:
@lazylier do you think your name should be put in "authors" or "others"

lazylier:
Change it however you think is best

Themostrandom:
image
Hello,

I find the spawn areas of the chests too dense, why not make them bigger?
The minerals are well distributed
. The 3:2 screen format is not respected.
I like the size of your map, finally a non speedrunnable map. 
However, I think there needs to be a minimum of 18-20 players.

AnonomousAccount:
From what i recall, the chest are all placed there because they looked really bad over snow (they would float over the snow) so binshaden cleared an area to put them.

AnonomousAccount:
> image Hello,
> 
> I find the spawn areas of the chests too dense, why not make them bigger? The minerals are well distributed . The 3:2 screen format is not respected. I like the size of your map, finally a non speedrunnable map. However, I think there needs to be a minimum of 18-20 players.

3:2 is the correct screenshot dimensions iirc?

a-blob:
@AnonomousAccount I don't think the current screenshot is in 3:2 ratio?

AnonomousAccount:
You're right, it wasn't, now it is though

AnonomousAccount:
New Map Version with new structures!

AnonomousAccount:
New map version! More buildings! (Bye bye barrier file)

gameit0:
There are a few thing that should be changed IMO:

The barrier is placed wrong/is set wrong in the config. The green team can enter blues' side of the map by 7 blocks. Blues can't move close to the barrier before being teleported back to their base.
screenshot_20250503_150934

There are massive mountains in the middle of the map, however there are no structures inside of them, not even natural caves. That could be changed to expand the variety of places to explore.

Players spawn with 99 torches. That's a bit much, isn't it? I mean, as soon as you dig one you need another slot of your inventory. IMO a lower amount like ~50 is more than enough, even if there isn't much natural light on the map – which is a nice feature.

I feel like the both bases are very different from each other. Blue's one is set on the ground and can reach their chest-zone pretty fast while green's one is far more protected by the water. However, they have a longer way to reach their chest-zone. Maybe there is a way to balance the bases a bit better.

There are no iron-ores at all on the map. Ores (not just mese/dias) should be there for classic-mode. They could probably be added somewhere in the mountains, along with some tunnels/mines and/or natural caves, like mentioned before.

Overall, it could be a nice map after some changes.

ClothierEdward:
Alright, I've finally got time to review this map so here's what I think about it. 
The map is very interesting with how massive it is compared to most maps. I think it might actually be the largest map in main CTF if it were to be added. Speaking of which, if it does get added, I hope the server can handle loading it in because when I booted up this map to test, it took a few minutes to load. 
I like how you added ladders and platforms to the mountain. It makes it considerably easier to traverse. It also adds a lot of a variety in terms of how you want to get past the mountain. I will say though, I wish there were more mineshafts, especially ones that went through the mountain. 
I also don't like how the bases are different sizes. Either make green base or make the blue base smaller. Other than that, I like where this map is going. It definitely has potential to be a challenging map, especially if the player count is high. I believe if you make the few changes I suggested, it would be a very good fit for main CTF.

PR #377, opened 2025-01-26
Add "The Upside down"

ClothierEdward:
This map is an April-Fools themed map that would be exclusive to the April Fools Event that happens every April. The map takes place in an upside down forest with two wooden shacks serving as the team bases. In order to traverse the map, players will have to either parkour through trees, bridge with blocks, or tunnel to make it to the other side. There are loot chests on the trees and there are loot chests you can find by digging straight up.


AnonomousAccount:
Is this on a new branch? I would suggest resetting your master branch and creating a new branch for each map.

ClothierEdward:
> Is this on a new branch? I would suggest resetting your master branch and creating a new branch for each map.

Yeah, I did make it on a new branch. It's on a different one from the one that had my Halloween map and Christmas map.

ClothierEdward:
> The safes seem to be buggy, I can't open them. I suggest changing the loots for less ordinary items and the same for the initial stuff. Otherwise, your map seems well suited to the 1st of April.

Yeah, I just tested the loot chests and yeah for some reason some of them can't open but I can't figure out why it does that. I did just update the loot for the map though. The loot chests now have steel, mese, diamond, and wool blocks.

PR #378, opened 2025-01-29
Add map Golden Grove Showdown

ScaleRaker:
This map is made by Dracophor. Please do not forward any thoughts on the map to me.


AnonomousAccount:
Missing Screenshot

Dracophor:
I will add a screenshot...

lazylier:
This map has nearly 300 distance between flags. That's probably too much. I think the flags should be moved farther away from the corners, maybe more centered along the sides too. The long distance and the lack of rear approaches could make it very difficult to capture. 

Aside from that, it would look much better to fill in the gap in the trees along the barrier, and also to have some smaller hills/valleys to break up the flatness of the grass. I do like the lumpy style of the mountains.

Dracophor:
its fine i dont care if my map gets added now i dont really want to sit and fix my map alot if it gets added cool if not cool i dont really care if you guys dont like it thats fine with me

ScaleRaker:
You will have to put a good amount more work into this project if you want to continue it. You should probably decide now whether to keep going or bag the project all together.

PR #310, opened 2023-08-15
BioSpheres by Kadax, TheCello, TheViola, and JackofLava

JackofLava:

PR #380, opened 2025-02-15
Add map Biospheres

KadaxCTF:


a-blob:
The .conf file is missing.

AnonomousAccount:
Continuation of https://github.com/MT-CTF/maps/pull/310 ?

a-blob:
Screenshot still needs to be in 3:2 ratio (screenshot 600x400px is suggested). You can just delete the screenshot from GitHub, crop it with tools such as https://imagy.app/image-aspect-ratio-changer/, and re-upload the cropped version.

KadaxCTF:
Alright I will try to do that when I access to a PC so it may be awhile but
thanks for letting me know.

KadaxCTF:
I have access*

KadaxCTF:
Ok I think I got the screenshot to the 3:2 ratio.

KadaxCTF:
Is this good or is there anything else I need to change/fix?

SaiPanneerselvam:
I think your map files may need work. When trying to play bio_spheres, i get the error message:
image

SaiPanneerselvam:
So again, the `map.conf` file is missing, making your map unplayable. The error:
image
shows that.

SaiPanneerselvam:
After applying a monkeypatch of sorts, it looks like there is only one team base, which is blue. (please correct me if i am wrong)
image
image
And there also is a random stone wall-like thing over in one of the biospheres

KadaxCTF:
Ok I'll try to fix the conf idk why it's not working now and idk how the heck that random walk got in there.

KadaxCTF:
OK I think I fixed the problems although when I played it I couldn't change my class does anybody know why that is?

SaiPanneerselvam:
1. Biospheres is spelled as BioSperes LOL. Probably you should fix that.
2. The hint isnt useful, remember that its optional.
3. It's extremely hard to know which is a base and which is just a biome, only team red's base is actually visible. 
4. Why is there a second map right below the normal one that is separated by ctf:ignores? (its not really a problem, just a question)

Overall, it looks like you rushed the creation of your map.conf. Otherwise, you seem all good.

KadaxCTF:
Sheesh everytime I touch it I screw something else up

KadaxCTF:
Also I like the bases how they are I don't think I'll change them.

a-blob:
@SaiPanneerselvam
> Why is there a second map right below the normal one that is separated by ctf:ignores? (its not really a problem, just a question)

Load the map in a new world every time and that shouldn't happen, unless the map export actually included another map.

KadaxCTF:
It does'nt have a second map under it when I load it so idk

a-blob:
Then it probably isn't your problem. When testing maps, if you load maps into an old world it might get placed over an already placed map.

KadaxCTF:
Whoops once again I touched it and screwed something up lol

SaiPanneerselvam:
Seems like it needs to be named differently
image

a-blob:
> Seems like it needs to be named differently
> image
> 

The screenshot is named screenshot.png so that's correct. Running the script in the scripts folder of the game should fix your error.

SaiPanneerselvam:
You are correct, that seemed to do just the thing. Thank!

SaiPanneerselvam:
So, Is this ready to be pushed/merged with the main?

KadaxCTF:
Is it all good then?

a-blob:
PR seems good, would be better if the screenshot shows more parts of the map since now a lot of it is cropped out.

KadaxCTF:
Ok if I get a moment I'll try to fix that.

KadaxCTF:
ok that should do it

a-blob:
Overall interesting maps, seems fun and well-built.
Here are a few problems I've noticed:
* Team zones seem to be set wrong on a few teams. On purple during build time, if you go to the middle of the map inside the middle sphere, you get teleported back for crossing the build-time barrier. On red, if you go behind and next to the glass wall of the map, you get teleported back too.
> Even standing on the edge of the team-zone will teleport the player back to base, so make sure to include the barriers within your team-zone too, or players will get teleported back to base if they touch the barriers during build-time
* In my opinion, blue seems to be over-defended with what's basically a prebuilt base, while green seems under-defended with a ruin around it that barely covers anything and requires more work to patch it up compared to others teams. Might be a bit unfair.
* Each team seems to have around the same amount of ores, but some are a bit off; might need some readjusting:
purple d:16 m:20 i:22 c:12
green: d:14 m:20 i:17 c:14
red: d:16 m:20 i:38 c:15
blue: d:15 m:20 i:35 c:12

A few suggestions about the map:
* Since the middle water sphere is unbreakable with only a door for each team, it might be a bit hard to capture? Depends, but players might just block off the whole entrance with team doors and blocks, making it very hard to enter (and bridging around the sphere seems hard too). So maybe make the entrance a bit bigger.
* Since the middle sphere has water, it would be nice to add some sand and kelp at the bottom to complete the "biospheres" theme.

KadaxCTF:
Thanks for the feedback I will try to fix everything sometime next week.
Also I love the idea of the sand and kelp in the water biosphere I wish I
had thought of that.

KadaxCTF:
Ok, I fixed the ores and I think the team borders. Also, I added sand and kelp to the middle biosphere and weakened the blue base.

KadaxCTF:
Cool thanks

KadaxCTF:
> @KadaxCTF It looks good. It may need to be adjusted depending on whether individual bases offer an unfair advantage/are more difficult to capture. One minor drawback is that the teams share the chest zones with two teams each, which can lead to uneven numbers per team. But I think it should work for now. So approved.

Sounds good thank you

PR #388, opened 2025-11-29
add Dark Halls map

lazylier:
I haven't been able to play test this, so it might not work right.


a-bad-dev:
Haven't fully tested it yet but here are a few issues:
1. The barrier is visible in the screenshot, remove the barriers before taking the screenshot.
2. Got an error involving chest zones while loading the map, and some nodes behaved like chests (see screenshot), the chest zone(s) need adjustment.

The map is unlike any other map I've seen (excepting maybe a mix of Water Academy's tunnels and the underworld's theme in Titanomachy: Aftermath) 

screenshot_20251130_001808

lazylier:
I tried saving again. It works now for me.

LuismiInTheZone:
The map is something completely new to me, so games could be interesting. However, since it is too dark, it is hard to play, so I'd recommend adding some more torches to the inventory just in case, but it's up to you. Another thing you could look into is the secret rooms or secret places of ores, it would be pretty helpful to add something that helps finding those. It's almost impossible to find them since there aren't any clues (maybe I missed something?), so, for example, since I see you used a lot of fireflies, you could use those as tips.

The map seems to have no errors, I'm not sure if I saw some torch around that you might have forgotten to remove, but since I'm not certain wait until I do a deeper review. The map is really large and complex, so it's difficult to review everything.

dark_halls/map.conf
```diff
@@ -0,0 +1,21 @@
...
+size = return {z=199,x=187,y=65}
+map_version = 3
+enabled = true
+name = Dark Halls
+phys_gravity = 1
+initial_stuff = return {"default:pick_steel","default:torch 1","default:cobble 30"}
+author = binshaden
+hint = Not everything is unbreakable.
+license = CC-BY-SA 4.0
+others = Other info
```
LuismiInTheZone: Also remove what appears in "others", leave it empty

